PR #22, opened 2023-02-13
Regex updated to work with the new JS code from fsrs4anki_scheduler_qt5.js

MNastri:


L-M-Sherlock:
The helper should be compatible with the old scheduler code. I plan to start the new scheduler version with `4.x.x` if we have a large modification.

L-M-Sherlock:
I prefer to release the new helper add-on before the new scheduler. Could you help me check the code? @MNastri

MNastri (commented):
`beq_version()` compares if given version1 is greater than or equal to version2. 

I'd change the name to `is_version_greater_than_or_equal_to` (I like long descriptive names) or `ge_version` (same [naming](https://docs.python.org/3/reference/datamodel.html#object.__ge__) used in python)

`uses_new_code()` don't need to do its assertion since the call to 'beq_version' already does that assertion

PR #47, opened 2023-03-10
Feat decks with shared parameters

MNastri:
This creates the decks from `const decks_with_shared_config` in the custom scheduler. [related PR](https://github.com/open-spaced-repetition/fsrs4anki/pull/185)

I was not planning on sharing this since I thought no one woud want this, but I hope it helps...

### DIFF
--- a/shared_config_decks.py
+++ b/shared_config_decks.py
@@ -57,7 +57,7 @@ def get_shared_config_name_and_decks(
     decks_as_list = []
     for decks in decks_as_string:
         deck = decks.lstrip("[").rstrip("]").split(",")
-        deck = [dd.strip().lstrip('"').rstrip('"') for dd in deck]
+        deck = [dd.strip().lstrip(""""'""").rstrip(""""'""") for dd in deck]
         deck = [dd for dd in deck if dd != ""]
         decks_as_list += (deck,)
     return config, decks_as_list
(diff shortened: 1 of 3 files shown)


L-M-Sherlock:
Could it work compatibly with the older versions? And I recommend packing these codes in another `.py` file because the `utils.py` is too complex now.

MNastri:
The code checks presence of the const and skips the new code when it's not present, so I believe this should work for both new and old code.

`has_const_decks_with_shared_config_in_code()` is used to make the check but now that you asked I remembered of `geq_version()` that we could use instead/in tandem with the current check.

Which approach you think is better?

MNastri:
I just realized that I didn't check the implications of changing deckParams from `const` to `var`.

I'm going out now and I'll check that after returning

MNastri:
Just reverted the change from `const` to `var` because it isn't needed

MNastri:
I added the tests I was using when writing the code. 
You can delete them, if you wish, after you've checked that they are correct, because I don't want to bloat the addon with them (unless you're okay with it).

MNastri:
> Could it work compatibly with the older versions? And I recommend packing these codes in another `.py` file because the `utils.py` is too complex now.

made another module for this feature! check it out when you have the chance

L-M-Sherlock:
Sorry for my delayed reply. I found that many users didn't understand how to configure deck-specific parameters. So decks with shared parameters would be more complicated to them. Maybe integrating to Anki and configuring FSRS parameters in deck options is the best solution. So I will close this PR.

PR #66, opened 2023-04-16
Treat rescheduled cards as good reviews &, do not reschedule manually scheduled cards

### LINKED ISSUE #64: Reschedule not working for manually scheduled cards
I set up free days to Sunday (today) and did a rescheduling of all cards. What surprised me is that it added even more cards to the review queue instead of emptying it.  

I noticed is that only cards from my “comprehension” decks that were rescheduled like that, the ones that were manually set by “Set due date” like the ones in the picture, not all of them though. Some were left untouched (the last picture). They all seem to have been rescheduled with an interval of 5 days, which is the graduation interval for that config. 

I did some testing and the ones getting the interval of 5 days hit the "if last_s is None:" line in reschedule.py. last_s is set by the second to last review (why not the last one?). If it was manual, rescheduler will treat it like a new card, because manual review kind has no button_chosen attribute, and it triggers the "# forget card" part of the logic.

I propose we treat manually rescheduled cards as review ones with rating set to 3 (good answer). Anki treats them like that.

image
image
image
image

tkozybski:
Resolves #64,

### DIFF
--- a/reschedule.py
+++ b/reschedule.py
@@ -245,7 +245,7 @@ def reschedule(did, recent=False):
                 good_ivl = max(hard_ivl + 1, good_ivl)
                 easy_ivl = max(good_ivl + 1, easy_ivl)
             if rating == 0:
-                new_ivl = int(revlogs[-1].interval / 86400)
+                new_ivl = int(revlogs[0].interval / 86400)
             else:
                 new_ivl = [again_ivl, hard_ivl, good_ivl, easy_ivl][last_rating - 1]
             offset = new_ivl - card.ivl
--- a/reschedule.py
+++ b/reschedule.py
@@ -162,6 +162,19 @@ def reschedule(did, recent=False):
                 last_s = s
                 rating = revlog.button_chosen
 
+                # Manually rescheduled last entry, user knows better, needs special handling
+                # Don't change the interval they set
+                if rating == 0 and revlog.review_kind == REVLOG_RESCHED and i == len(revlogs)-1:
+                    if last_date is None:
+                        d = fsrs.init_difficulty(3)
+                        s = fsrs.init_stability(3)
+                        
+                    ivl = revlog.interval 
+                    r = math.pow(0.9, ivl / s)
+                    d = fsrs.next_difficulty(d, 3)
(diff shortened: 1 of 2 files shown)


L-M-Sherlock:
There are two ways to use Set Due Date: 
grafik
The first sets the interval property of the card (interval-changing), the second is like burying (interval-preserving).
We should make clear which we are talking about.

# 1. interval-changing Set Due Date

This is intended to be a way to rate a card after review where the user is able to fine-tune the interval.

If I remember correctly, the interval-changing Set Due Date also calculates whether the interval corresponds to the interval of a hard, good or easy grade. Or at least it used to do that when Set Due Date was not yet natively implemented in Anki but provided by the [ReMemorize add-on](https://web.archive.org/web/20211201021611/https://ankiweb.net/shared/info/323586997). For example:

> previous interval: 5
> Set Due Date: set interval to 25
> → Review log should treat this as an easy rating, because the new ivl is 500% of the old interval.

> previous interval: 5
> Set Due Date: set interval to 6
> → Review log should treat this as hard rating, because the new ivl is 120% of the old ivl.

And when I write "review log", I actually mean the FSRS4Anki Scheduler and Helper (for the purposes of the present discussion).

[edit: For clarification, I agree that "the user knows better" and FSRS4Anki Helper should not reschedule cards that have a pending manually set due date. My point is just that FSRS4Anki would optimally treat interval-changing manually set due dates (e.g. `100!`) in a more sophisticated way: not as 3 without regard to the interval but as a rating of 2, 3, or 4 depending on the ivl increase.]

# 2. interval-preserving Set Due Date 

Since this is just long-term burying, it should not be considered as a review/rating (since the review log doesn't record the normal `bury` action either). And since FSRS4Anki Helper leaves buried cards alone, it should also leave alone cards that have a pending interval-preserving Set Due Date.

_Originally posted by @galantra in https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/64#issuecomment-1510485067_
            
Could you response these cases?

L-M-Sherlock:
I still want to skip manual schedule due to https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/64#issuecomment-1529036161, so I close this PR.

PR #136, opened 2023-06-15
Fix/filter out cram revlog

L-M-Sherlock:


L-M-Sherlock:
Patch: [fsrs4anki-helper.zip](https://github.com/open-spaced-repetition/fsrs4anki-helper/files/11758637/fsrs4anki-helper.zip)

Issue #198, opened 2023-08-20
[Enhancement] Consider (re)learning cards in retention stats


user1823:
> The original true retention add-on decides whether to include a card in the stats or not based on whether it is a learning card or a review card. But, in my opinion, there is no difference between a learning card having a 3-day interval and a review card having a 3-day interval.
> 
> So, wouldn't it be better to filter cards based on their interval (such that all cards with intervals more than or equal to 1 day are included) rather than filtering them based on whether they are learning cards or review cards.

_Originally posted by @user1823 in https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/170#issuecomment-1671779335_

The importance of the implementing the above suggestion is evident in the following screenshots.

<img width="460" alt="image" src="https://github.com/open-spaced-repetition/fsrs4anki-helper/assets/92206575/c2c77a78-8276-4144-ab98-36d91589f61f"><img width="506" alt="image" src="https://github.com/open-spaced-repetition/fsrs4anki-helper/assets/92206575/8f382715-9e58-466b-94a3-251fec7d0a50">

Before I switched to FSRS, a majority of my daily reviews consisted of (re)learning cards because I used long (re)learning steps. If the helper add-on doesn't show these reviews, the result obtained is misleading.

By the way, there is a typo in the word "Retention" at three places in the above screenshot.

L-M-Sherlock:
How about only considering the (re)learning cards with interval longer or equal to one day?

L-M-Sherlock:
Patch: [fsrs4anki-helper.zip](https://github.com/open-spaced-repetition/fsrs4anki-helper/files/12395262/fsrs4anki-helper.zip)

Have a try.

user1823:
> Patch: [fsrs4anki-helper.zip](https://github.com/open-spaced-repetition/fsrs4anki-helper/files/12395262/fsrs4anki-helper.zip)

This patch didn't fix the problem completely. The result:
<img width="499" alt="image" src="https://github.com/open-spaced-repetition/fsrs4anki/assets/92206575/38652491-08a9-4f10-a87b-842cf33bbb3e">

Then, I made a further change, which seems to fix the problem. The result:
<img width="504" alt="image" src="https://github.com/open-spaced-repetition/fsrs4anki-helper/assets/92206575/1f13190f-a9db-46fc-be75-da006f6ead38">

I have made a PR with the above change. Please review.

[...]

PR #199, opened 2023-08-21
Fix/Consider (re)learning cards in retention stats

user1823:
Fixes https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/198


L-M-Sherlock (approved):

L-M-Sherlock:
Thanks! It works well.

Issue #194, opened 2023-08-19
[Feature Request] Option to reschedule card after review


user1823:
I have seen comments from many users that they don't want to reschedule all their cards while switching to FSRS from SM-2 because it results in a huge backlog.
Examples:
- https://www.reddit.com/r/Anki/comments/15rmp13/comment/jwa92er
- https://www.reddit.com/r/Anki/comments/15rmp13/comment/jwlc1nb

The current suggestion for them is to either not reschedule at all (and allow the FSRS scheduler to do all the work) or to use the "reschedule cards reviewed in last x days" option every day.

A better solution would be to add an option that reschedules the card as soon as it is reviewed. This would be quite similar to how the "disperse siblings after review" feature currently works.
For better results, this new feature should be executed before the disperse siblings feature so that the dispersal of the siblings is not reversed.

Expertium:
I think we've talked about something similar in the past, and I don't remember why this wasn't implemented. But yes, it would be a good feature for a more slow and seamless transition from Anki's algorithm to FSRS.

L-M-Sherlock:
Patch: [fsrs4anki-helper.zip](https://github.com/open-spaced-repetition/fsrs4anki-helper/files/12406380/fsrs4anki-helper.zip)
Have a try.

Expertium:
![image](https://github.com/open-spaced-repetition/fsrs4anki-helper/assets/83031600/ea5975be-c15a-42d0-ae3f-18086f3db7d1)
I don't see any changes.

L-M-Sherlock:
I haven't added the config. But it has worked in all reviews.

L-M-Sherlock:
Do you need me to add tooltip to inform you that the card has been rescheduled after review?

Expertium:
I think that it should become a new option, like "Auto reschedule the card after review". The issue is that users will ask "Wait, but then why do we even need the scheduler if the helper add-on is doing this?".

L-M-Sherlock:
Oh, the patch has a bug. Here is the fixed version with tooltip (for test).
[fsrs4anki-helper.zip](https://github.com/open-spaced-repetition/fsrs4anki-helper/files/12407481/fsrs4anki-helper.zip)


> I think that it should become a new option, like "Auto reschedule the card after review". The issue is that users will ask "Wait, but then why do we even need the scheduler if the helper add-on is doing this?".

Yeah. So how to explain it to the user?

Expertium:
[...]

PR #202, opened 2023-08-21
Fix card counts in true retention table

user1823:
The previous commit (https://github.com/open-spaced-repetition/fsrs4anki-helper/commit/83fec94aaa9aac327c2276bf2f5e53825b84268b) had some problems.

- With multi-day (re)learning steps, it counted all the (re)learning steps. This caused the number of (re)learned cards to be spuriously high.
- When a learning card was flunked during its learning steps, it was counted again in the learned cards. However, it should have been ideally counted in the relearned cards.

This PR aims to solve both of the above problems.

Before:
image

After the 1st commit (which solves the first problem above):
image

After the 2nd commit (which solves the second problem above):
image

After the 4th commit (which solves the problem below):
image

~~However, unfortunately, the first commit seems to filter out more than the desired number of reviews. For example, the number of cards graduated in the past week decreased from 114 to 42. But my (re)learning steps are shorter than 1d since Feb 2023. So, I don't know why it is behaving in this manner.~~

Edit:
Ok, I found the problem. The cards which didn't get "Again" or "Hard" as the first rating would have first interval ≥ 1d. So, my first commit will remove those cards as well. I have fixed this in the fourth commit.

In the latest screenshot, the total number of graduated cards (29645) is slightly greater than the number of review cards (`-is:new`) in my deck (28869). I don't know why. If you can find it out and push a fix, it would be great. Otherwise, I think that this branch could be merged and a new issue be created for that discrepancy so that we can address it later.

### DIFF
--- a/stats.py
+++ b/stats.py
@@ -245,8 +245,8 @@ def get_true_retention(self):
                 <td class="trc" colspan=3><span class="young"><b>Young</b></span></td>
                 <td class="trc" colspan=3><span class="mature"><b>Mature</b></span></td>
                 <td class="trc" colspan=3><span class="yam"><b>Young and Mature</b></span></td>
-                <td class="trc" rowspan=2><span class="young"><b>Learnt</b></span></td>
-                <td class="trc" rowspan=2><span class="relearn"><b>Relearnt</b></span></td>
+                <td class="trc" rowspan=2><span class="young"><b>Learned</b></span></td>
+                <td class="trc" rowspan=2><span class="relearn"><b>Relearned</b></span></td>
             </tr>
             <tr>
                 <td class="trc"><span class="young">Pass</span></td>
--- a/stats.py
+++ b/stats.py
@@ -278,8 +278,8 @@ def stats_list(lim, span):
     sum(case when lastIvl < %(i)d and ease > 1 and (type = 1 OR lastIvl <= -86400 OR lastIvl >= 1) then 1 else 0 end), /* passed young */
     sum(case when lastIvl >= %(i)d and ease = 1 and (type = 1 OR lastIvl <= -86400 OR lastIvl >= 1) then 1 else 0 end), /* flunked mature */
     sum(case when lastIvl >= %(i)d and ease > 1 and (type = 1 OR lastIvl <= -86400 OR lastIvl >= 1) then 1 else 0 end), /* passed mature */
-    sum(case when (ivl >= 1 OR ivl <= -86400) and type = 0 then 1 else 0 end), /* learned */
-    sum(case when (ivl >= 1 OR ivl <= -86400) and type = 2 then 1 else 0 end) /* relearned */
+    sum(case when (ivl >= 1 OR ivl <= -86400) and (lastIvl > -86400 and lastIvl < 0) and type = 0 then 1 else 0 end), /* learned */
+    sum(case when (ivl >= 1 OR ivl <= -86400) and (lastIvl > -86400 and lastIvl < 0) and type = 2 then 1 else 0 end) /* relearned */
     from revlog where id > ? """ % dict(i=config.mature_ivl) + lim, span)
     yflunked, mflunked = yflunked or 0, mflunked or 0
     ypassed, mpassed = ypassed or 0, mpassed or 0
--- a/stats.py
+++ b/stats.py
@@ -278,8 +278,8 @@ def stats_list(lim, span):
     sum(case when lastIvl < %(i)d and ease > 1 and (type = 1 OR lastIvl <= -86400 OR lastIvl >= 1) then 1 else 0 end), /* passed young */
     sum(case when lastIvl >= %(i)d and ease = 1 and (type = 1 OR lastIvl <= -86400 OR lastIvl >= 1) then 1 else 0 end), /* flunked mature */
     sum(case when lastIvl >= %(i)d and ease > 1 and (type = 1 OR lastIvl <= -86400 OR lastIvl >= 1) then 1 else 0 end), /* passed mature */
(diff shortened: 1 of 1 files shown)


L-M-Sherlock:
Please read this document: https://github.com/ankidroid/Anki-Android/wiki/Database-Structure#review-log

And what's your definition for the number of graduated cards and relearned cards?

user1823:
> Please read this document: https://github.com/ankidroid/Anki-Android/wiki/Database-Structure#review-log

I have read it. It was indeed a great resource for implementing the above.

> And what's your definition for the number of graduated cards and relearned cards?

- Graduated card is any card that gets an interval ≥ 1d for the first time. However, I think that it could be renamed as "Learned" card.
- Relearned card is any card that gets failed after getting an interval ≥ 1d. If a card is failed multiple times, it is counted multiple times.

If you need my opinion on something, I wouldn't be able to comment before tomorrow evening. Sorry for that.

L-M-Sherlock:
> * Graduated card is any card that gets an interval ≥ 1d for the first time. However, I think that it could be renamed as "Learned" card.

`the first time` is problem. A card could include multiple review logs with type = 0 and ivl >= 1, but your SQL query doesn't consider it.

> * it is counted multiple times.

It's inconsistent with `Graduated card`. If they both are named xxx `cards`, why the fronter are only counted once but the latter are counted multiple times?

user1823:
> the first time is problem. A card could include multiple review logs with type = 0 and ivl >= 1, but your SQL query doesn't consider it.

Can you provide an example other than the ones that I have already accounted for? I considered two cases:
- In cards with multi-day learning steps, the previous code counted all the learning steps with ivl > 1d. However, after https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/202/commits/8ce3c2555c9b42d4116282cd82a096efeab78dbc, only those learning steps are counted where the last interval was shorter than 1d. This approach also had a problem that it didn't count those cards whose first interval was ≥ 1d. This was fixed by https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/202/commits/9e2c71143178ca3c84fa0294f3a84e94cadb79ea.
- The other possibility was that a card was rated `Again` during its learning steps after it had got an ivl ≥ 1d. So, it would be counted again. But, this was fixed by https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/202/commits/60f46ce63cbba63945c08eb40b701fa14fa1801d so that anytime a card is failed during its learning steps (with ivl ≥ 1d), it is subtracted from the total count.

> > * it is counted multiple times.
> 
> It's inconsistent with `Graduated card`. If they both are named xxx `cards`, why the former are only counted once but the latter are counted multiple times?

This is consistent with the behaviour of the original True Retention add-on. If you consider a collection with all (re)learning steps shorter than 1d, the original True Retention add-on would count Graduated cards only once but Relearned cards as many times as they get failed.

L-M-Sherlock:
Maybe I need to check your database. I checked the database of @Experium:

image

There revlogs are counted multiple times.

L-M-Sherlock:
I write a query that could solve that problem. You can test it.
```sql
SELECT 
COUNT(DISTINCT case when type = 0 and (ivl >= 1 OR ivl <= -86400) and (lastIvl > -86400 and lastIvl <= 0) then cid else NULL end)
FROM revlog;
```

user1823:
> Maybe I need to check your database.

You can use the .apkg that I shared previously. The results should be replicable in that because I haven't used multi-day learning steps since then.

L-M-Sherlock:
> You can use the .apkg that I shared previously. The results should be replicable in that because I haven't used multi-day learning steps since then.

image

You can check these cards' info:

1625652147161
1645806382820
1645870735802
1645878277821
1645891951985

L-M-Sherlock:
> The first card has a manual reset in the learning phase. What SQL query can I use to find these? The Database Structure page linked above doesn't talk about manual ratings. I think that not accounting these cards is the main problem.

Anki's database changes and the documents rot over time. So it is very hard to understand it fully. Here is my query:

```
SELECT *
FROM revlog
WHERE cid in (
SELECT cid
FROM (
SELECT *
FROM revlog
WHERE type = 0 and (ivl >= 1 OR ivl <= -86400) and (lastIvl > -86400 and lastIvl <= 0))
GROUP BY cid
HAVING COUNT(*) > 1)
AND   type = 0 and (ivl >= 1 OR ivl <= -86400) and (lastIvl > -86400 and lastIvl <= 0)
;
```

L-M-Sherlock:
This query can help you find out the extra cards:
```sql
select   *
from revlog 
where cid in 
(select id 
from cards )
AND type = 0 and (ivl >= 1 OR ivl <= -86400) and (lastIvl > -86400 and lastIvl <= 0)
AND cid NOT in (
SELECT id
FROM cards
WHERE type = 0)
;
```

L-M-Sherlock:
Oh, I figure it out. `-is:new` also filter out manual reset cards.

user1823:
> Oh, I figure it out. -is:new also filter out manual reset cards.

Do you mean to say that if a card has manual reset but it is not new (i.e. it has been reviewed after being reset), it would be filtered out by using `-is:new`? If so, I think that it is a bug on the part of Anki.

L-M-Sherlock:
> Do you mean to say that if a card has manual reset but it is not new (i.e. it has been reviewed after being reset), it would be filtered out by using `-is:new`? If so, I think that it is a bug on the part of Anki.

I mean a card has manual reset and it's new, then `-is:new` will filtered out it. But our sql will count it.

L-M-Sherlock:
Do you think this card should be counted in graduated card?

image

user1823:
> we can concat this condition into the query:

I added this condition. The result was:
image

Then, I removed the condition `(lastIvl > -86400 and lastIvl <= 0)` because its purpose had already been solved by using DISTINCT and adding too many conditions may lead to unexpected behavior. The result was:
image

However, the count unfortunately fell below the expected number. If you don't have any further ideas, I would vote for merging this branch because it is already a lot better than the current main and small discrepancies can be dealt with later.

L-M-Sherlock (approved):

L-M-Sherlock:
Thanks. I will update in AnkiWeb after #194 has been merged.

PR #210, opened 2023-08-28
Fix/correct formula for new_ivl in disperse_siblings

user1823:
In the following code, an incorrect formula for `new_ivl` would be used if the user didn't remove the value of `easy bonus` from their scheduler code after updating to v4.
https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/857cba9a70adf62473c07ebc7b8f17f7d416facb/schedule/disperse_siblings.py#L49-L54

I fixed this in the current PR. However, I am not very sure about the correctness of the syntax of the code. So, please verify.

Also, in the above code, the `new_ivl` was calculated using the old formula in v4 when last rating ≠ 4. I have corrected this too.

### DIFF
--- a/schedule/disperse_siblings.py
+++ b/schedule/disperse_siblings.py
@@ -300,14 +300,14 @@ def disperse_siblings_when_review(reviewer, card: Card, ease, undo_entry=None):
     if len(siblings) <= 1:
         return
     
-    # messages = []
+    messages = []
     
     card_cnt = 0
     undo_entry = mw.col.add_custom_undo_entry("Disperse") if undo_entry is None else undo_entry
     best_due_dates = disperse(siblings)
     for cid, due in best_due_dates.items():
         card = mw.col.get_card(cid)
-        # old_due = card.odue if card.odid else card.due
+        old_due = card.odue if card.odid else card.due
         last_revlog = mw.col.card_stats_data(cid).revlog[0]
         last_due = get_last_review_date(last_revlog)
         card = update_card_due_ivl(card, last_revlog, due - last_due)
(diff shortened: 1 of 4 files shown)


schedule/disperse_siblings.py
```diff
@@ -47,13 +47,13 @@ def get_due_range(cid, parameters, stability, due):
     revlogs = filter_revlogs(mw.col.card_stats_data(cid).revlog)
     last_due = get_last_review_date(revlogs[0])
     last_rating = revlogs[0].button_chosen
-    if last_rating == 4:
-        try:
-            new_ivl = int(round(stability * parameters['e'] * math.log(parameters['r']) / math.log(0.9)))
-        except KeyError:
-            new_ivl = int(round(9 * stability * (1 / parameters['r'] - 1)))
+    if version[0] == 4:
```
L-M-Sherlock, line 50: We need pass the version variable when calling this function.

L-M-Sherlock (changes requested):

schedule/disperse_siblings.py
L-M-Sherlock: Version is not a global variable.

user1823: So, how to do it?

user1823:
With the latest commit, I have renamed "Young and Mature" in the true retention table to "Total" for conciseness.

schedule/disperse_siblings.py
L-M-Sherlock: Have you tested it in your device? I will review it tomorrow.

user1823: I have not tested it because I don't exactly how to test this.

L-M-Sherlock: You can replace the code in your local add-on file.

user1823: > You can replace the code in your local add-on file.

I knew that. The problem was that I didn't know what to look while testing. But, then I decided to test it anyway. I initially got an error and then fixed it. Apart from that, it seems to work fine (but I am not sure).

L-M-Sherlock (approved):

PR #228, opened 2023-09-11
Feat/Scheduler updater

Luc-Mcgrady:
Allows you to update the scheduler code from within Anki.

image

### Without scheduler:
image

### With updated scheduler:
image

### With out of date scheduler:

image
image
image

Not sure how needed this is with the inbuilt rust on the way but whatever.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .DS_Store
 .idea/
 .vscode/
+__pycache__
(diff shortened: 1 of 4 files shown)


Luc-Mcgrady:
That should do it.

user1823:
If the installed scheduler is v3, we should advise them to re-optimize their parameters (using v4 optimizer) and then provide a solution for them to update their scheduler code to v4.

L-M-Sherlock (approved):
LGTM. I add timeout to avoid freezing Anki.

user1823:
> If the installed scheduler is v3, we should advise the user to re-optimize their parameters (using v4 optimizer) and then provide a solution for them to update their scheduler code to v4.

@Luc-Mcgrady, can you make another PR to implement this?

PR #230, opened 2023-09-11
Feat/Update from V3 option

Luc-Mcgrady:
As requested by @user1823 in #228 

image

Did you want something like this? It shows whenever you update with a V3 scheduler.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .DS_Store
 .idea/
 .vscode/
+__pycache__
--- a/__init__.py
+++ b/__init__.py
@@ -9,6 +9,7 @@ from .schedule.postpone import postpone
 from .schedule.advance import advance
 from .schedule.reset import reset
 from .schedule.disperse_siblings import disperse_siblings
+from .schedule.update_scheduler import update_scheduler
 from .schedule import init_review_hook
 from .stats import init_stats
 from .browser.browser import init_browser
@@ -118,6 +119,8 @@ menu_reset = build_action(reset, "Undo reschedulings in all cards")
 
 menu_disperse_siblings = build_action(disperse_siblings, "Disperse all siblings")
 
+menu_update_scheduler = build_action(update_scheduler, "Update scheduler")
+
 menu_for_helper = mw.form.menuTools.addMenu("FSRS4Anki Helper")
 menu_for_helper.addAction(menu_auto_reschedule)
 menu_for_helper.addAction(menu_auto_reschedule_after_review)
@@ -133,6 +136,7 @@ menu_for_helper.addAction(menu_postpone)
 menu_for_helper.addAction(menu_advance)
 menu_for_helper.addAction(menu_reset)
 menu_for_helper.addAction(menu_disperse_siblings)
+menu_for_helper.addAction(menu_update_scheduler)
 
 
 def set_free_days(day, checked):
--- a/schedule/update_scheduler.py
+++ b/schedule/update_scheduler.py
@@ -7,7 +7,7 @@ import socket
 import re
 import os.path
 
-scheduler_qt_suffix = "_qt5" if qVersion().split(".")[0] == 5 else ""
+scheduler_qt_suffix = "_qt5" if qVersion().split(".")[0] == "5" else ""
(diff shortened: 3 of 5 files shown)


user1823:
Yes, I asked for something like this.

But, the things are slightly more complicated. The config in the v4 scheduler code doesn't contain easy bonus and hard interval. So, these should be removed while updating from v3 to v4.

user1823:
Not directly related, but could you make this preserve the custom scheduling code that is placed above the FSRS scheduler. Some users use additional custom scheduling (for minor things) in addition to the FSRS scheduler.

Luc-Mcgrady:
### Before v3 upgrade
image
### After
image

I hope this doesn't give people the wrong impression that the updater updates their weights to be suitable as well.
Anything else?

user1823:
> It is recommended that you upgrade to the V4 scheduler as the V3 scheduler is no longer getting updated.
> This will require you to re-optimize your decks.

This can be changed to:

> FSRS4Anki Helper has detected that you are using the v3 FSRS scheduler. The v4 FSRS scheduler has been released and the v3 FSRS scheduler is no longer getting updates. So, it is recommended that you upgrade to the v4 FSRS scheduler.
> 
> NOTE: The weights of the v3 FSRS scheduler are incompatible with those of the v4 FSRS scheduler. So, if you choose to upgrade to v4, you must re-optimize your weights using the optimizer and then replace the weights in the scheduler config. Here is the link to the latest optimizer: https://colab.research.google.com/github/open-spaced-repetition/fsrs4anki/blob/main/fsrs4anki_optimizer.ipynb

image

In my opinion, this warning is not important while updating from v3 to v4 because we have already warned the user in the previous message.

image

When updating from v3 to v4, this screen can also be used to remind the user to re-optimize their weights.

> FSRS scheduler updated from v3 to v4 successfully. Do remember to re-optimize your weights using the optimizer and then replace the weights in the scheduler config.

image

Replace "local" with "installed" (many users won't understand what "local" means).

user1823:
LGTM

L-M-Sherlock:
A problem needs to resolve: the scheduler has two different versions: Q5 and Q6. We need to check the version before we update it.

Luc-Mcgrady:
I can not test that as I don't have QT5 Anki installed but that should work.

user1823:
Another issue (it is becoming too complicated 😅):

If a user is not using all the deck configurations, they will not update the parameters of those configurations, leading to something like this:
https://github.com/open-spaced-repetition/fsrs4anki/issues/458#issuecomment-1715883567

Solution:
When updating from v3 to v4, replace all parameters with the default v4 parameters (but don't change the deck names & request retention). This won't increase the work of the users because the older parameters are not valid anyway.

Luc-Mcgrady:
Like that?!?
(This took me way longer than it feels like it should have)

user1823:
This replaces the requestRetention too, but I advised you not to do so because the user's target retention for the decks would not change when the user upgrades from v3 to v4.

> Your deck param configs will each be replaced with the default until you do this.

Remove this. I have an alternative in mind (see below).

image

When updating from v3 to v4, remove the part that says "Your config will be preserved" because it is only partially true and can be misleading.

image

Change it to:

> FSRS scheduler updated from v3 to v4 successfully.
> 
> The weights of the v3 and the v4 FSRS scheduler are incompatible with each other. So, to avoid errors, the weights have been replaced by the default weights for v4.
> 
> For better results, re-optimize your weights using the optimizer and then manually replace the weights in the scheduler config.


By the way, thanks for working on this. When this gets merged, it would make it much easier for the users to upgrade from v3 to v4.

Luc-Mcgrady:
image

Eyy it doesnt look as cluttered as I thought it would.
> (I figure we shouldn't put the ***full*** re-optimization warning here as we already told them 2 screens ago)

I did (kind of) mean the first one but I must not have edited the comment in time, I'm just worried about hitting people with walls of text.

Glad I could help anyway.

user1823:
LGTM

user1823 (approved):

L-M-Sherlock (approved):

PR #234, opened 2023-09-15
Retention graph by Card Maturity

user1823:
The retention graph in the FSRS Stats can be misleading because it doesn't separate low stability and high stability reviews.

For many users, SM-2 tends to show low interval reviews at an unnecessarily high frequency causing the retention to go up. If the retention for the young and mature cards is not separated, this effect can mask the low retention in the high interval reviews.

So, it makes sense to separate young and mature cards in the retention graph.

The True Retention table fixes this to some extent, but it shows the cumulative retention of all the reviews till today and not the retention at a particular point of time. So, it is somewhat difficult to interpret.

Retention graph using current version of the add-on:
image

Retention graph after my changes (currently bugged):
image

Although it is not showing the curve for Mature cards, but you can see that the retention in the Young cards is clearly more than the total retention.

I have more improvements to make. But, firstly I need some help in getting the curve for mature cards displayed.

### DIFF
--- a/stats.py
+++ b/stats.py
@@ -198,6 +198,9 @@ def get_retention_graph(self: CollectionStats):
 
     rate_data_young, _, rate_data_mature, _, cnt_data, _ = data
 
+    recall_min = min(min(item[1], item[2]) for item in offset_retention_review_cnt)
+    recall_max = max(max(item[1], item[2]) for item in offset_retention_review_cnt)
+
     rate_data_young["lines"] = {"show": True}
     rate_data_young["bars"] = {"show": False}
     rate_data_young["yaxis"] = 1
@@ -218,8 +221,8 @@ def get_retention_graph(self: CollectionStats):
         xaxis=dict(tickDecimals=0, max=0.5),
         yaxes=[
             dict(
-                min=0,
-                max=1,
+                min=recall_min,
+                max=recall_max,
                 ticks=[[x / 10, str(round(x / 10, 1))] for x in range(0, 11)],
             ),
             dict(position="right", min=0),
--- a/stats.py
+++ b/stats.py
@@ -172,7 +172,8 @@ def get_retention_graph(self):
 
     query = f"""SELECT
     CAST((id/1000.0 - {mw.col.sched.day_cutoff}) / 86400.0 as int)/{chunk} AS day,
-    SUM(CASE WHEN  ease == 1 THEN 0.0 ELSE 1.0 END) / COUNT(*) AS retention,
+    SUM(CASE WHEN ease == 1 AND lastIvl < 21 THEN 0.0 ELSE 1.0 END) / COUNT(*) AS retention_young,
+    SUM(CASE WHEN ease == 1 AND lastIvl >= 21 THEN 0.0 ELSE 1.0 END) / COUNT(*) AS retention_mature,
     COUNT(*) AS review_cnt
     FROM revlog
     WHERE (type = 1 OR lastIvl <= -86400 OR lastIvl >= 1)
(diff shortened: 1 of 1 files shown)


L-M-Sherlock:
image

How about that?

user1823:
I tried to correct the count used in the denominator. But, I don't know why this change caused the following error to appear. I didn't change the number of output values.

> Caught exception:
> Traceback (most recent call last):
>   File "aqt.webview", line 46, in cmd
>   File "aqt.webview", line 153, in _onCmd
>   File "aqt.webview", line 665, in _onBridgeCmd
>   File "aqt.toolbar", line 411, in _linkHandler
>   File "aqt.toolbar", line 432, in _statsLinkHandler
>   File "aqt.main", line 1227, in onStats
>   File "aqt", line 130, in open
>   File "aqt.stats", line 178, in __init__
>   File "aqt.stats", line 224, in refresh
>   File "anki.stats", line 113, in report
>   File "C:\Users\user\AppData\Roaming\Anki2\addons21\759844606\stats.py", line 93, in todayStats_new
>     + get_retention_graph(self)
>   File "C:\Users\user\AppData\Roaming\Anki2\addons21\759844606\stats.py", line 191, in get_retention_graph
>     rate_data_young, _, rate_data_mature, _, cnt_data, _ = data
> ValueError: not enough values to unpack (expected 6, got 4)

In addition to finding a fix for this, I need help with
- replacing lastIvl in line 169 and 170 by "elapsed days"
- replacing 21 d in line 169 and 170 by mature_ivl from config
- making the scale of the y-axis such that there is a larger space available for the retention range 0.8 to 1.0. An even better solution would be to make this adjustable based on the value of the retention.

L-M-Sherlock:
> But, I don't know why this change caused the following error to appear.

Fixed.

> * replacing lastIvl in line 169 and 170 by "elapsed days"

It's tricky, because the revlog doesn't record it.

> * replacing 21 d in line 169 and 170 by mature_ivl from config

Done.

> * making the scale of the y-axis such that there is a larger space available for the retention range 0.8 to 1.0. An even better solution would be to make this adjustable based on the value of the retention.

Do you mean that the floor and ceiling should depend on the min/max of the retention?

user1823:
> > But, I don't know why this change caused the following error to appear.
>
> Fixed.

So, the error was due to the case where the denominator became 0. Right?

> > * replacing lastIvl in line 169 and 170 by "elapsed days"
> 
> It's tricky, because the revlog doesn't record it.

Yes, it is tricky. But, I think that it can be calculated using a code similar to line 33 to 36.

> > * making the scale of the y-axis such that there is a larger space available for the retention range 0.8 to 1.0. An even better solution would be to make this adjustable based on the value of the retention.
> 
> Do you mean that the floor and ceiling should depend on the min/max of the retention?

Yes, but the floor of the graph should be above the bars representing the count. Otherwise, the two graphs (retention and count) will overlap.

user1823:
@Expertium, do you have any suggestions on this (retention graph by card maturity)?

Currently, the retention graph looks like this for my collection:

image

L-M-Sherlock:
> So, the error was due to the case where the denominator became 0. Right?

Yeah.

> Yes, it is tricky. But, I think that it can be calculated using a code similar to line 33 to 36.

They are different. Line 33 to 36 is used for cards table. It only calculates the latest elapsed days of cards.

Expertium:
I agree that y-axis values should be adjusted based on min and max values of retention.

L-M-Sherlock:
image

user1823:
Sorry for bothering you again and again. But most of the ideas that are coming to my mind currently can be implemented only by adding new lines of codes (instead of tweaking the existing ones). So, I can't implement them on my own.

Ideas:
- Instead of using the exact min and max values of retention as the floor and the ceiling, we can use the following approach:
    - Make list of the numbers 0, 0.1, 0.2, 0.3, … 0.8, 0.9 and 1.
    - Now, floor is the number from the above list that is just smaller than the min value of retention.
    - Similarly, ceiling is the number from the above list that is just larger than the max value of retention.
- The distance between two consecutive labels on the y-axis can be taken as `(ceiling - floor)/10`.
- Move the bars behind the lines just like the [graph I shared earlier](https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/234#issuecomment-1722237181).
- Separate the `review_cnt` based on maturity (just like retention) and display it as a stacked bar chart (like in the following graph).
image


image

L-M-Sherlock:
image

user1823:
In my case, the retention rate has only three labels (0.8, 0.9 and 1.0). I think that it should have more labels.

image

Probably, the distance between two consecutive labels on the y-axis should be taken as `(ceiling - floor)/10` or `(ceiling - floor)/5`.

L-M-Sherlock:
image

user1823:
I tried to make some changes so that we need to calculate the number of reviews once instead of twice so as to improve performance. However, I now get this error.

> Caught exception:
> Traceback (most recent call last):
>   File "aqt.webview", line 46, in cmd
>   File "aqt.webview", line 153, in _onCmd
>   File "aqt.webview", line 665, in _onBridgeCmd
>   File "aqt.toolbar", line 411, in _linkHandler
>   File "aqt.toolbar", line 432, in _statsLinkHandler
>   File "aqt.main", line 1227, in onStats
>   File "aqt", line 130, in open
>   File "aqt.stats", line 178, in __init__
>   File "aqt.stats", line 224, in refresh
>   File "anki.stats", line 113, in report
>   File "C:\Users\user\AppData\Roaming\Anki2\addons21\759844606\stats.py", line 95, in todayStats_new
>     + get_retention_graph(self)
>   File "C:\Users\user\AppData\Roaming\Anki2\addons21\759844606\stats.py", line 200, in get_retention_graph
>     data, _ = self._splitRepData(
>   File "anki.stats", line 473, in _splitRepData
> IndexError: tuple index out of range
>

L-M-Sherlock:
I don't think it could improve the performance. And the old stats code is very hard to maintain. I recommend reverting the latest commit and merge this branch.

user1823:
Ok, I have (partially) reverted the change and made some more changes.

image

L-M-Sherlock (approved):

PR #237, opened 2023-09-18
Correct due ranges in disperse

user1823:


schedule/disperse_siblings.py
```diff
@@ -70,28 +70,30 @@ def get_due_range(cid, parameters, stability, due):
...
 
     last_elapsed_days = (
         int((revlogs[0].time - revlogs[1].time) / 86400) if len(revlogs) >= 2 else 0
     )
     min_ivl, max_ivl = get_fuzz_range(new_ivl, last_elapsed_days)
     if due >= mw.col.sched.today:
-        due_range = (max(last_due + min_ivl, due), max(last_due + max_ivl, due), cid)
+        due_range = (max(last_review + min_ivl, mw.col.sched.today), max(last_review + max_ivl, mw.col.sched.today), cid)
+    elif last_review + max_ivl > mw.col.sched.today:
+        due_range = (mw.col.sched.today, last_review + max_ivl, cid)
```
L-M-Sherlock, line 82: If due is greater than today, the due_range[0] shouldn't be equal to today. I think `max(last_review + min_ivl, mw.col.sched.today)` is not better than max(last_due + min_ivl, due).

If due is less than today, and last_review + max_ivl is greater than today, why should due_range[0] be greater than today?

user1823: > If due is greater than today, the due_range[0] shouldn't be equal to today.

The due is greater than today, not the main value of fuzz range. So, even if the card becomes due today, it will still remain in the fuzz range. (Here, by fuzz range, I mean the original fuzz range that was obtained in line 78).

> I think max(last_review + min_ivl, mw.col.sched.today) is not better than max(last_due + min_ivl, due).

The problem with max(last_due + min_ivl, due) is that if the current `due` is near the maximum limit of the fuzz range, we are unnecessarily shortening the `due_range`. Why should it not be allowed to take `due` values before the current `due`?

> If due is less than today, and last_review + max_ivl is greater than today, why should due_range[0] be greater than today?

Allowing the card to take `due` values before today doesn't make any sense because the user will still have to do it today or in future (the user can't do it the past by going back in time).

L-M-Sherlock (approved):

PR #238, opened 2023-09-19
Refactor/native FSRS support

L-M-Sherlock:
Support Anki 23.10+


user1823:
Here, you are removing the support for Anki versions older than 2.1.67. However, it doesn't seem to be a good idea to support _only_ Anki 2.1.67+.

Either you should make a new add-on for Anki 2.1.67+ or you should update the add-on in such a way that it supports the current versions of Anki as well as 2.1.67+

By the way, Anki is changing its versioning system. So, the next version is going to be called 2023.10.0 (and not 2.1.67)

L-M-Sherlock:
> Here, you are removing the support for Anki versions older than 2.1.67. However, it doesn't seem to be a good idea to support _only_ Anki 2.1.67+.

The old version is still available. LIke this:

image

schedule/advance.py
```diff
@@ -73,31 +57,19 @@ def advance(did):
         lambda x: (
             x
             + [
-                did_to_deck_parameters[x[1]]["r"],
-                exponential_forgetting_curve(x[4], x[3])
-                if version[0] == 3
-                else power_forgetting_curve(x[4], x[3]),
+                DM.config_dict_for_deck_id(x[1])["desiredRetention"],
```
dae, line 60: This is now available via $.dr

L-M-Sherlock: What is `$.dr`?

dae: The desired retention is stored in each card to make sorting easier, so you don't need to fetch it from the deck preset.

user1823:
Does the reschedule function still need to have its own code for calculating the memory states from the revlog? If FSRS code is embedded into Anki, I think that the add-on should be able to call the backend code to calculate the memory states.

L-M-Sherlock:
> Does the reschedule function still need to have its own code for calculating the memory states from the revlog? If the FSRS code is embedded into Anki, I think that the add-on should be able to call the backend code to calculate the memory states.

I plan to remove the code of memory states calculation, because Anki updates all cards' memory states when the user modify the weights.

L-M-Sherlock:
But for cards reviewed in other device, it's necessary to use the add-on's own code to re-calculate them.

L-M-Sherlock:
> I don't think so. I think that the add-on can call the backend code for calculating the memory states for such cards too.
> 
> If this is not currently possible, I think a very small change to the backend code (adding a hook) might enable this.

Yeah. dae will implement it in a few days.

dae:
Very minor: native, not naive

schedule/reschedule.py
```diff
@@ -298,88 +221,20 @@ def reschedule_background(did, recent=False, filter_flag=False, filtered_cids={}
...
-            last_rating = rating
-
-    if rating is None or s is None:
-        return None
+def reschedule_card(cid, fsrs: FSRS, recompute=False):
+    card = mw.col.get_card(cid)
+    if recompute:
+        memory_state = mw.col.compute_memory_state(cid)
+        s = memory_state.stability
+        d = memory_state.difficulty
```
dae, line 229: These will be None in the case of new cards, so initial d/s still needs to be dealt with.

L-M-Sherlock: I have exclude new cards with:

```
        WHERE queue IN ({QUEUE_TYPE_REV}, {QUEUE_TYPE_REV}, {QUEUE_TYPE_DAY_LEARN_RELEARN})
```

L-M-Sherlock:
@dae, do you have plan to integrate Advance/Postpone/Load Balance/Free days/Disperse Siblings into Anki? By the way, I find that Reschedule has been implemented now, so I will remove it in the add-on.

user1823:
@dae, I find that the native Reschedule option in Anki adds a `Manual` entry to the revlog.
- Is it intended/desirable?
- Is the FSRS code able to handle this entry when re-training?

image

dae:
It might be a bit early to remove your reschedule implementation - the one in the Rust code is integrated into deck options changes, so can't be called directly, and as you've seen, it writes an entry into the revlog, so it's not something you'd want to be calling every day. I would rather users upgrade all of their clients so they don't need to reschedule every day, as it creates extra work for AnkiWeb when syncing. Once 23.10 is released as stable (or is close), I'll get an update sent to AnkiDroid and ask them to put out an alpha.

Because I'd like 23.10 to come out as stable this month if possible, I don't think any of the others will make this release. I would like to integrate Free Days in the future, and maybe advance/postpone/load balancing at a later date (some thought will be required to ensure they are efficient, and don't result in the entire collection's cards being synced each time the user uses them). I'm not so sure about sibling dispersal, and will need to give that more thought.

The manual entry is ignored when training or computing memory states, and is ignored by the standard Anki stats screen - it's only the legacy stats screen that can't cope with it.

dae:
> Is it intended/desirable?

It serves a similar purpose to the v=reschedule line that helper adds - it identifies why a card's interval/due date doesn't match the earlier review log.

user1823:
This version doesn't seem to disperse all the siblings after sync.

After sync, I got the message "Rescheduled 72 cards. Dispersed 2 siblings in 1 note." (not exactly this, based on memory)

But, I reviewed many cards having siblings. I had expected something like "Dispersed 40 siblings in 20 notes".

user1823:
I feel that the add-on is filtering out buried cards when searching for siblings. This is a problem because Anki automatically buries siblings when answering a card.

L-M-Sherlock:
Here is the code to find siblings:

https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/565701a91c5278f59e2a1e11afd8dbe666ea83d4/schedule/disperse_siblings.py#L22-L49

user1823:
But, there should be something that is preventing the siblings from being dispersed.

user1823:
Another somewhat minor issue is that the tooltip saying "x cards rescheduled" is not always shown after sync even though reschedule after sync is enabled.

user1823:
@L-M-Sherlock, did you try finding out why is auto-disperse not working as intended?

user1823:
> This version doesn't seem to disperse all the siblings after sync.

@L-M-Sherlock, please debug this. I am still facing this problem.

Issue #243, opened 2023-09-23
[Feature Request] Update the README


Expertium:
https://www.reddit.com/r/Anki/comments/16pg5au/comment/k1qr4yz/?utm_source=share&utm_medium=web2x&context=3
See the comment and the post. The user was confused about "Auto reschedule the card after each review", "Update scheduler", "Reschedule all cards" and "Reschedule cards reviewed in last 7 days" options. @user1823, it would be great if you could help.

PR #244, opened 2023-09-23
Doc/Explain about more features of the add-on

user1823:
Fixes https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/243

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # FSRS4Anki Helper
 
-FSRS4Anki Helper is an Anki add-on that supports [FSRS4Anki](https://github.com/open-spaced-repetition/fsrs4anki) scheduler. It has six features:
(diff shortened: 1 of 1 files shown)


L-M-Sherlock:
@Expertium, could you help me review this PR?

Expertium (commented):
Change this:
"It might have a small effect on responsiveness and power consumption because it needs more number of calculations for each review and causes constant queue rebuilding."
To this:
"It might have a small effect on responsiveness of Anki and introduce lag because FSRS needs to perform more calculations for each review, which causes constant queue rebuilding."

user1823:
Updated

L-M-Sherlock (approved):

L-M-Sherlock:
Thanks for this contribution. I will update the description in the add-on page tomorrow.

PR #248, opened 2023-09-26
Memory states from SM-2 for incomplete review history

user1823:
Context: https://github.com/open-spaced-repetition/fsrs-rs/issues/63#issuecomment-1735323658

I think that I have added most of the code required to make it work. You will just need to do some final touches so that it can become functional. Specifically, you will need to edit line 391 and maybe minor changes to the other parts.

I hope that editing this PR requires less effort than writing all the code yourself.

### DIFF
--- a/schedule/reschedule.py
+++ b/schedule/reschedule.py
@@ -398,8 +398,13 @@ def reschedule_card(cid, fsrs: FSRS, rollover, params):
     reps = len(revlogs)
     for i, revlog in enumerate(reversed(revlogs)):
         rating = revlog.button_chosen
-        if i == 0 and revlog.review_kind == REVLOG_REV:
-            s, d = fsrs.memory_state_from_sm2(revlog.ease/1000, revlog.interval/86400, retention)
+        if i == 0 and (
+            revlog.review_kind == REVLOG_REV
+            or (revlog.review_kind == REVLOG_CRAM and revlog.ease != 0)
+        ):
+            s, d = fsrs.memory_state_from_sm2(
+                revlog.ease / 1000, revlog.interval / 86400, retention
+            )
             s = max(0.1, min(s, 36500))
             d = constrain_difficulty(d)
             again_s = s
@@ -448,7 +453,7 @@ def reschedule_card(cid, fsrs: FSRS, rollover, params):
                 datetime.fromtimestamp(revlog.time - rollover * 60 * 60).toordinal()
                 - last_date.toordinal()
             )
-            if elapsed_days <= 0 and revlog.review_kind in (REVLOG_LRN, REVLOG_RELRN):
+            if elapsed_days <= 0:
                 continue
             r = (
                 exponential_forgetting_curve(elapsed_days, s)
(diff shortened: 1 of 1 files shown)


L-M-Sherlock:
Patch: [fsrs4anki-helper.zip](https://github.com/open-spaced-repetition/fsrs4anki-helper/files/12729635/fsrs4anki-helper.zip)

Could you help me test it?

user1823:
It didn't work, i.e., the card didn't get custom data.

For example, this card:

L-M-Sherlock:
Patch: [fsrs4anki-helper.zip](https://github.com/open-spaced-repetition/fsrs4anki-helper/files/12729854/fsrs4anki-helper.zip)

I hope it solve the problem.

user1823:
It seems to work well, though I will test it more.

However, I am slightly curious as to why the interval decreased from 1.04 years to 1.01 years. However, this is not a problem.

image

L-M-Sherlock:
> However, I am slightly curious as to why the interval decreased from 1.04 years to 1.01 years. However, this is not a problem.

It would be due to fuzz.

user1823:
I have tested this and it seems to be working fine. Thanks!

I am attaching some screenshots for future reference (or if you are interested in seeing them).

image
image
image
image

user1823:
Before (left) After (right)
imageimage

Total Count increased by 87 cards. 🚀🚀🚀

user1823:
Before merging this, I advise you to check how this affects cards with the first rating as "Manual". Examples:


Source of images: https://forums.ankiweb.net/t/anki-23-10-beta/34912/18

I can't test this case because I don't have such cards.

user1823:
Also, we should test for the case where first review kind is FILTERED and "reschedule cards based on my answers in this deck" is enabled & disabled respectively.

L-M-Sherlock:
> Also, we should test for the case where first review kind is FILTERED and "reschedule cards based on my answers in this deck" is enabled & disabled (both cases separately).

If you enable "reschedule cards based on my answers in this deck", the review kind is REVIEW instead of FILTERED.

> Before merging this, I advise you to check how this affects cards with the first rating as "Manual". Examples:

The second review is treated as the first learning.

user1823:
> > Also, we should test for the case where first review kind is FILTERED and "reschedule cards based on my answers in this deck" is enabled & disabled (both cases separately).
> 
> If you enable "reschedule cards based on my answers in this deck", the review kind is REVIEW instead of FILTERED.

It doesn't seem to be the case when the card is reviewed before it is due.

image


> > Before merging this, I advise you to check how this affects cards with the first rating as "Manual". Examples:
> 
> The second review is treated as the first learning.

Assuming that this is actually what happens, it looks good to me.

L-M-Sherlock:
> It doesn't seem to be the case when the card is reviewed before it is due.

Weird. I read the doc of Anki, it said that:

    /// Old Anki versions called this "Cram" or "Early", and assigned it when
    /// reviewing cards ahead. It is now only used for filtered decks with
    /// rescheduling disabled.

Source: https://github.com/ankitects/anki/blob/05499297e065b7536e701e105eb613a761b7c135/rslib/src/revlog/mod.rs#L65-L75

I need to ask @dae.

dae:
That comment is wrong - it's also applied when a review card is answered early (eg not using is:due to create the filtered deck)

L-M-Sherlock:
Fine. Could I distinguish these two types of review via ease != 0?

dae:
Yes, I think so - the v2 scheduler didn't write revlog entries when not rescheduling, and v3 uses ease 0 when not rescheduling.

PR #262, opened 2023-10-08
Improve fuzz seed calculation

user1823:
Apply left_rotate to cid as in https://github.com/ankitects/anki/blob/23f29a6eccfa762f76f1a9926e44f8983a1cc565/rslib/src/card/mod.rs#L253-L257

The rotation code is based on https://stackoverflow.com/a/49462383 and https://www.geeksforgeeks.org/rotate-digits-of-a-given-number-by-k/.

### DIFF
--- a/schedule/disperse_siblings.py
+++ b/schedule/disperse_siblings.py
@@ -77,7 +77,11 @@ def get_due_range(cid, parameters, stability, due):
     )
     min_ivl, max_ivl = get_fuzz_range(new_ivl, last_elapsed_days)
     if due >= mw.col.sched.today:
-        due_range = (max(last_review + min_ivl, mw.col.sched.today), max(last_review + max_ivl, mw.col.sched.today), cid)
+        due_range = (
+            max(last_review + min_ivl, mw.col.sched.today),
+            max(last_review + max_ivl, mw.col.sched.today),
+            cid,
+        )
     elif last_review + max_ivl > mw.col.sched.today:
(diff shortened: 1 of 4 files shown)


user1823:
I think that the fuzz seed is re-calculated in Anki after every review because it depends upon `reps`. However, it is not so in the helper add-on.

In my opinion, this is because the add-on needs to ensure that if a card is rescheduled twice without any review (of the same card) in between, the fuzz seed should remain the same. So, the add-on reuses the fuzz seed from the custom data (if available).

The problem with this approach is that some cards are always scheduled near the lower limit of the fuzz range while others are always scheduled near the upper limit of the fuzz range. This decreases the effectiveness of fuzz.

So, is there a solution to this problem? Can the situation be made better in the native implementation of FSRS?

L-M-Sherlock:
> The problem with this approach is that some cards are always scheduled near the lower limit of the fuzz range while others are always scheduled near the upper limit of the fuzz range. This decreases the effectiveness of fuzz.

I think this problem doesn't exist. The seed is changed in the custom scheduling during each review.

user1823:
> I think this problem doesn't exist. The seed is changed in the custom scheduling during each review.

Oh, I see. But, in my opinion, this problem exists for people who are not using the scheduler.js (e.g. people who do their reviews on AnkiDroid). However, I think that this issue is not a priority.

L-M-Sherlock (approved):

PR #264, opened 2023-10-08
Fix/has no attribute 'fuzz factor'

user1823:
@L-M-Sherlock, your fix in https://github.com/open-spaced-repetition/fsrs4anki-helper/commit/e64371ce344215fb42502a53988f596cd537962b helped me to locate the problem and work out a possibly better fix.

My version ensures that set_fuzz_factor is not called when the seed is already there in the old custom data. It saves 4 seconds on my collection when rescheduling (25 s vs 21 s).


L-M-Sherlock (approved):
Good job!

Issue #265, opened 2023-10-13
[BUG] helper throwing an error while starting Anki 23.10beta5: ValueError: Protocol message Column has no "sorting" field.


szalejot:
**Describe the bug**
At the start of 23.10beta5:

Error
An add-on you installed failed to load. If problems persist, please go to the Tools>Add-ons menu, and disable or delete the add-on.

When loading '⁨FSRS4Anki Helper⁩':
⁨Traceback (most recent call last):
  File "aqt.addons", line 239, in loadAddons
  File "C:\Users\michal\AppData\Roaming\Anki2\addons21\759844606\__init__.py", line 14, in <module>
    from .browser.browser import init_browser
  File "C:\Users\michal\AppData\Roaming\Anki2\addons21\759844606\browser\browser.py", line 9, in <module>
    from .custom_columns import (
  File "C:\Users\michal\AppData\Roaming\Anki2\addons21\759844606\browser\custom_columns.py", line 50, in <module>
    class TargetRetrievabilityColumn(CustomColumn):
  File "C:\Users\michal\AppData\Roaming\Anki2\addons21\759844606\browser\custom_columns.py", line 51, in TargetRetrievabilityColumn
    builtin_column = Column(
ValueError: Protocol message Column has no "sorting" field.

**To Reproduce**
1. Install Anki 23.10beta5 version.
2. Open.
3. See error

**Expected behavior**
No error.

**Environment**
 - Anki version: 23.10beta5
 - OS: Windows 10

**Additional context**

dae:
.sorting has become .sortingNotes / .sortingCards - it should be .SORTING_NONE in the case of .sortingNotes.

L-M-Sherlock:
@user1823, @Expertium

Patch: [fsrs4anki-helper.zip](https://github.com/open-spaced-repetition/fsrs4anki-helper/files/12902730/fsrs4anki-helper.zip)

PR #268, opened 2023-10-13
Temporary workaround for https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/265

user1823:
Approach: Disable browser features


L-M-Sherlock:
#270 has fixed #265

Issue #259, opened 2023-10-07
[Feature Request] Change "Disperse Siblings" so that it cannot increase the number of cards due today


Expertium:
**Which module is related to your feature request?**
Scheduler and Helper.

**Is your feature request related to a problem? Please describe.**
Right now using Disperse Siblings can increase the number of cards due today. This is unintuitive (when people see the word "disperse" they assume it will _decrease_ the number of reviews) and discourages people from using this feature.

**Describe the solution you'd like**
Suppose there are 2 cards that are siblings. There are 3 possible scenarios:
1) Both of them are due later, not today. In that case, dispersing either of them should not affect the number of cards due today.
2) One of them is due today, the second is due later. In that case, the interval of one of them needs to increase, which, again, shouldn't increase how many cards are due today. If the first card is postponed, the number of cards due today will decrease by 1. If the second card is postponed, the number of cards due today shouldn't change.
3) Both of them are due today. In this case, one of them should be dispersed, therefore decreasing the number of due cards today by 1.

**Additional context**
I'm using the latest version of the Helper and 23.10beta4.

user1823:
> Both of them are due later, not today. In that case, dispersing either of them should not affect the number of cards due today.

What if both of them are due on the same day (or close to each other)? In this case, one (or both) of the cards needs to be dispersed. What if the maximum distance is achieved by scheduling one of the cards today?

Here is the relevant code that decides which dates the function can access while dispersing the siblings.

https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/c5cad0017f93858498831ea97442b750708371ac/schedule/disperse_siblings.py#L78-L85

Do you have any specific suggestions for the code?

Expertium:
> What if both of them are due on the same day (or close to each other)? In this case, one (or both) of the cards needs to be dispersed. What if the maximum distance is achieved by scheduling one of the cards today?

The interval of one of those cards should become longer. The key idea is to ensure that Disperse Siblings **cannot make intervals shorter**. @L-M-Sherlock, I don't know how exactly to implement it, but I hope you agree with what I said. If Disperse Siblings can make intervals of cards shorter, it can result in the number of due cards increasing after using this feature, which is undesirable, as it confuses users and gives them more work.

L-M-Sherlock:
> The key idea is to ensure that Disperse Siblings **cannot make intervals shorter**.

But it would also make the retention lower and lower.

Expertium:
[...]

PR #287, opened 2023-10-28
+1 for due when min_due = 0 in disperse siblings

user1823:
In https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/259#issuecomment-1783808514, I didn't suggest the following change.

image

Reason: It affects all the cards, even those that were ideally dispersed before. 

My change only affects the cards where min_gap = 0.

By the way, I am sure that there is some bug in my code (for e.g., are the following lines valid?) Your help in making it functional would be appreciated.

```py
    if min_gap == 0:
        for due in best_due_dates.items():
```

### DIFF
--- a/schedule/disperse_siblings.py
+++ b/schedule/disperse_siblings.py
@@ -130,11 +130,6 @@ def disperse(siblings):
     due_ranges[-1] = (latest_review, latest_review)
     min_gap, best_due_dates = maximize_siblings_due_gap(due_ranges)
     best_due_dates.pop(-1)
-    
-    if min_gap == 0:
-        for due in best_due_dates.items():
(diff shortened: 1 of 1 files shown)


user1823:
Ideally, the 1 should be added for all **except** one card. But, I am not sure how to do it? So, I added it to all cards instead.

L-M-Sherlock:
I think there is a better method: due = max(due, today+1).

L-M-Sherlock:
I recommend adding the code in disperse after review instead of disperse(). Because the later is also called in disperse all siblings.

user1823:
My last commit assumes that disperse_siblings_when_review doesn't affect (re)learning cards.

If it's not the case, we can revert the last commit.

PR #304, opened 2023-11-10
Slight changes to the "Overview" section.md

Expertium:
Question: how does Load balancing work with the built-in version of FSRS, considering that Anki has it's own fuzz factor? I'm worried that Load Balancing + Anki's fuzz factor can result into something weird. Also, am I correct when I say that rescheduling using the add-on is no longer necessary?


L-M-Sherlock:
Anki built-in fuzz factor is irrelevant when load balance is enabled.

PR #328, opened 2023-12-18
Feat/auto apply easy days when open collection

L-M-Sherlock:


__init__.py
```diff
@@ -158,7 +160,11 @@ def set_easy_days(day, checked):
 menu_for_easy_4 = checkable(title="Easy Fri", on_click=lambda x: set_easy_days(4, x))
 menu_for_easy_5 = checkable(title="Easy Sat", on_click=lambda x: set_easy_days(5, x))
 menu_for_easy_6 = checkable(title="Easy Sun", on_click=lambda x: set_easy_days(6, x))
+menu_for_auto_easy_days = checkable(
+    title="Auto easy days when open collection", on_click=lambda x: set_auto_easy_days(x)
```
user1823, line 164:
```suggestion
    title="Auto apply easy days on opening collection", on_click=lambda x: set_auto_easy_days(x)
```

user1823:
I have not tested whether it works correctly or not (because I don't use easy days). But, the title was confusing. So, I made a suggestion to improve it.

__init__.py
jonahspear-toast: I like this grammar better as well.

jonahspear-toast:
Just tested the patch you send me and works great! Thank you so much.

My one request would be that we do this behavior on close rather than on open. Why? I think it would be better for end users.

Imagine that I am going for a trip and I Study 100 cards and then close my laptop. There will still be many cards scheduled on the weekend. If I take my phone with me only on that trip, then I will have to study on the weekend. To get these cards rescheduled, I would have to close anki, open it again (to reschedule cards), then close it one more time. This seems like a messy/painful user experience.

rescheduling cards on close instead would fix that. 

I think this small fix would make this perfect. If you don't have time

L-M-Sherlock:
> I think this small fix would make this perfect. If you don't have time, I'm also happy to help with it (although I would need some resources about how to build/test this application)

PR is welcome. Resources: https://addon-docs.ankiweb.net/hooks-and-filters.html

PR #335, opened 2023-12-27
Use new forgetting curve in disperse_siblings

user1823:


L-M-Sherlock (approved):
Thanks! I missed that.

PR #342, opened 2024-01-06
Feat/reschedule selected cards in browser

L-M-Sherlock:


schedule/reschedule.py
```diff
@@ -244,3 +246,16 @@ def reschedule_card(cid, fsrs: FSRS, recompute=False):
...
+
+
+def reschedule_browser_selected_cards(browser: browser.Browser):
+    cids = browser.selected_cards()
+    reschedule(None, False, True, cids)
+
+
+@browser_menus_did_init.append
+def on_browser_menus_did_init(browser: browser.Browser):
+    action = QAction("FSRS: update memory state & reschedule", browser)
```
user1823, line 258:
```suggestion
    action = QAction("FSRS: Update memory state and reschedule", browser)
```

user1823: The `&` icon is not correctly displayed.

user1823:
The patch works correctly. I have left a comment regarding incorrect rendering of the icon.

Issue #324, opened 2023-12-14
[Feature Request] Let's settle the disagreement regarding "Disperse Siblings" with the community


Expertium:
**Which module is related to your feature request?**
Helper add-on

**Is your feature request related to a problem? Please describe.**
Right now, Disperse Siblings can bring siblings **closer together** in some cases, in other words, the distance between their due dates (in days) can decrease. This is undesirable for two reasons:
1) It confuses the users since they expect that siblings will only become spaced further away from each other once this feature is enabled.
2) It defeats the point of this feature, which is to amplify the spacing effect for all siblings.

**Describe the solution you'd like**
Ensure that Disperse Siblings can only keep the distance (in days) between two cards the same or increase it, but cannot decrease it.

**Additional context**
Me and @L-M-Sherlock have been arguing about this for quite some time now, so I decided to make a Google Form survey. I will post it on r/Anki, if Sherlock isn't against it. We will let all FSRS users (well, those of them who use Reddit, I guess) vote on how to handle this.
@user1823 @AnKingMed @Dae https://forms.gle/a8CyrYK9EpZYFszSA

user1823:
In my opinion, the only case where the current implementation decreases the gap is the case when this decrease is necessary to increase the gap from the last review date (which is equally important in my opinion).

Expertium:
You can fill out my survey and choose "Other" in the last question.

user1823:
The way you have written about the issue, no one will select the second or third option (unless they know why the add-on can sometimes decrease the spacing).

Expertium:
Then, would you make a better write-up? Preferably, one that explains why the current implementation should remain unchanged in layman's terms.

user1823:
The Disperse Siblings feature is designed to maximize
- the gap between the due dates of the siblings; and
- the gap between the latest review date of the siblings and the current due dates of the siblings. 

The function also ensures that the due dates of all the siblings stay within their respective fuzz ranges, preventing any card from diverging too much from its optimal interval. This is crucial for maintaining the balance between sibling dispersion and retention.

Occasionally, the function can bring the due dates of siblings closer together rather than dispersing them. This occurs in an attempt to increase the gap from the latest review date, which is equally important. 

The issue of siblings being brought closer could theoretically be solved by allowing dispersion beyond their fuzz ranges. However, this approach introduces several potential problems:

[...]

PR #350, opened 2024-01-24
Prevent disperse siblings from modifying "overdue" cards

user1823:
https://www.reddit.com/r/Anki/comments/19e0x2l/comment/kjapnru/

If the users actually want to schedule the cards according to the FSRS algorithm, they should use `Reschedule`. Disperse siblings should not have the same effect as reschedule.

### DIFF
--- a/schedule/disperse_siblings.py
+++ b/schedule/disperse_siblings.py
@@ -101,7 +101,7 @@ def get_due_range(cid, stability, due, desired_retention, maximum_interval):
         int((revlogs[0].time - revlogs[1].time) / 86400) if len(revlogs) >= 2 else 0
     )
     min_ivl, max_ivl = get_fuzz_range(new_ivl, last_elapsed_days, maximum_interval)
-    if due > mw.col.sched.today and last_review + max_ivl < mw.col.sched.today:
+    if due > last_review + max_ivl:
         due_range = (due, due)
     elif due >= mw.col.sched.today:
         due_range = (
--- a/schedule/disperse_siblings.py
+++ b/schedule/disperse_siblings.py
@@ -101,7 +101,9 @@ def get_due_range(cid, stability, due, desired_retention, maximum_interval):
         int((revlogs[0].time - revlogs[1].time) / 86400) if len(revlogs) >= 2 else 0
     )
     min_ivl, max_ivl = get_fuzz_range(new_ivl, last_elapsed_days, maximum_interval)
-    if due >= mw.col.sched.today:
+    if due > mw.col.sched.today and last_review + max_ivl < mw.col.sched.today:
+        due_range = (due, due)
+    elif due >= mw.col.sched.today:
         due_range = (
             max(last_review + min_ivl, mw.col.sched.today),
             max(last_review + max_ivl, mw.col.sched.today),


L-M-Sherlock:
If so, should we calculate the fuzz range based on stability? Why not use the card's elapsed days?

user1823:
> If so, should we calculate the fuzz range based on stability? Why not use the card's elapsed days?

Can you elaborate?

L-M-Sherlock:
I misunderstand the "overdue". What do you mean the "overdue"? Why we shouldn't change the due when `due > last_review + max_ivl`?

user1823:
Basically, the due date of the card is beyond the maximum limit of the fuzz range. So, it most likely means that the user has not rescheduled their cards after updating FSRS weights (or after first switching to FSRS from SM-2).

If the user chose not to reschedule the cards, disperse siblings should not force it.

There can be other causes (e.g. bugs) of `due` being greater than the maximum limit, but disperse siblings doesn't have the responsibility of correcting this issue. The issue can be corrected by using `Reschedule` before using `Disperse`.

> I misunderstand the "overdue". 

Sorry for the terminology. It makes more sense when you consider my first commit. In that, rescheduling the cards would have made them overdue (due before today). I got the term from there and didn't change it after I made the rule more general.

user1823:
@L-M-Sherlock, what's your opinion?

This change is very simple, and I think that it can solve https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/324 too.

Expertium may be having the issue because he doesn't reschedule cards after changing the parameters. I have never experienced this issue and that's probably because I always reschedule all cards after changing the parameters.

L-M-Sherlock:
Fine. It will be a temporary workaround until I figure out the new method to disperse siblings.

L-M-Sherlock (approved):

user1823:
> It will be a temporary workaround

I don't think that it is temporary, i.e., I don't think that this should be removed at any point of time.

L-M-Sherlock:
If so, why not just check the last revlog entry to check whether the card has been manual rescheduled?

user1823:
> why not just check the last revlog entry to check whether the card has been manual rescheduled? 

What if the cards was rescheduled with
- Anki's built-in option
- helper add-on

Code to cover both these cases would be of several lines.

Also, how do you ensure that the card was not rescheduled?  Does the interval in the revlog need to be exactly the same as that in the upper section of the card info? What if the user had used disperse siblings in the past too? In that case, the intervals would be different (but still below the maximum limit).

In short, I think that my approach is a more efficient way of dealing with this.

> we don't know whether the user don't want to reschedule a card.

If they want to reschedule, they should press `Reschedule`. If they don't press Reschedule, we assume that they don't want to do it (most likely because of the backlog).

L-M-Sherlock:
I get it, but there is another problem:

> If they want to reschedule, they should press `Reschedule`. If they don't press Reschedule, we assume that they don't want to do it (most likely because of the backlog).

They don't want to reschedule doesn't mean they don't want to disperse siblings. But the current method doesn't allow the card to disperse because the due range is fixed in (due, due).

user1823:
Ok, I will make another PR.

PR #351, opened 2024-01-25
Disperse siblings - Create a new fuzz range for cards scheduled beyond max fuzz limit

user1823:
Follow up of #350

Specifically, to solve:

The current method doesn't allow the card to disperse because the due range is fixed in (due, due).

_Originally posted by @L-M-Sherlock in https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/350#issuecomment-1909275123_


L-M-Sherlock (approved):
Thanks, LGTM

user1823:
Can you update the add-on on AnkiWeb today itself so that users can take advantage of the better handling of non-rescheduled cards?

PR #353, opened 2024-01-25
Feat/apply Easy Days to more than one date

L-M-Sherlock:
image

Patch: [fsrs4anki-helper.ankiaddon.zip](https://github.com/open-spaced-repetition/fsrs4anki-helper/files/14048948/fsrs4anki-helper.ankiaddon.zip)

@user1823, give it a try.

### DIFF
--- a/schedule/easy_days.py
+++ b/schedule/easy_days.py
@@ -2,6 +2,7 @@ from aqt import (
     QDate,
     QDateEdit,
     QDateTime,
+    QHBoxLayout,
     QLabel,
     QPushButton,
     QVBoxLayout,
@@ -61,10 +62,11 @@ def auto_easy_days():
 
 
 # Modified from https://github.com/sam1penny/countdown-to-events/blob/main/src/__init__.py#L86-L169
-class EasySpecificDateWidget(QWidget):
+class EasySpecificDateManagerWidget(QWidget):
     def __init__(self):
         super().__init__()
         self.initUi()
+        self.specific_dates = []
 
     def initUi(self):
         self.layout = QVBoxLayout()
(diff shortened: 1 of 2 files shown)


user1823:
If I want to apply easy days to a single date, with this implementation, I have to first click `Add` and then `Apply`.
In my opinion, clicking `Apply` should run the function using the date entered within the box and `Add` should be used to add more dates.

user1823:
Also, can you change the date format in the input box to ISO 8601 standard (YYYY-MM-DD)? 
I can't change the date to anything less than 25 unless I change the month first. (Anki is trying to prevent me from entering past dates.) In the ISO standard, the month comes first. So, the user will change the month first and the issue will not arise. In addition, the ISO standard is less confusing.

L-M-Sherlock:
> In my opinion, clicking `Apply` should run the function using the date entered within the box and `Add` should be used to add more dates.

I'm afraid that it would induce confusion. Some users likely don't know the date in the DateEditBox will also be taken into account.

> Also, can you change the date format in the input box to ISO 8601 standard (YYYY-MM-DD)?

Do you mean a text input box? I don't want to deal with invalid user inputs.

> Otherwise, you can remove the following line and add a check after the user has pressed `Add` or `Apply`.

OK. I prefer this way.

L-M-Sherlock:
patch: [fsrs4anki-helper.ankiaddon.zip](https://github.com/open-spaced-repetition/fsrs4anki-helper/files/14049384/fsrs4anki-helper.ankiaddon.zip)

schedule/easy_days.py
```diff
@@ -73,32 +75,50 @@ def initUi(self):
         self.dateLabel.setText("Select The Specific Date")
```
user1823:
```suggestion
        self.dateLabel.setText("Select the Date(s)")
```

schedule/easy_days.py
```diff
@@ -73,32 +75,50 @@ def initUi(self):
...
         self.dateEdit = QDateEdit()
         self.dateEdit.setDateTime(QDateTime.currentDateTime())
-        self.dateEdit.setMinimumDate(QDate.currentDate())
 
-        self.applyEasyDayBtn = QPushButton("Apply Easy Day")
+        self.addDateBtn = QPushButton("Add the Selected Specific Date")
```
user1823, line 79:
```suggestion
        self.addDateBtn = QPushButton("Add the Selected Date")
```

schedule/easy_days.py
```diff
@@ -73,32 +75,50 @@ def initUi(self):
...
             tooltip("Please enable load balance first")
             return
-        specific_date = self.dateEdit.date().toPyDate()
-        current_date = datetime.now().date()
-        day_offset = (specific_date - current_date).days
-        today = mw.col.sched.today
-        specific_due = today + day_offset
-        due_days = [specific_due]
+        if len(self.specific_dates) == 0:
+            tooltip("Please add specific dates first")
```
user1823, line 112:
```suggestion
            tooltip("Please add the dates first.")
```

schedule/easy_days.py
```diff
@@ -61,44 +62,63 @@ def auto_easy_days():
...
         self.setLayout(self.layout)
-        self.setWindowTitle("Easy Day for a specific date")
+        self.setWindowTitle("Easy Days for Specific Dates")
+
+    def addEventFunc(self):
+        specific_date = self.dateEdit.date().toPyDate()
+        if specific_date in self.specific_dates:
+            tooltip("This date has already been added")
+            return
+        self.specific_dates.append(specific_date)
```
user1823, line 100:
```suggestion
        if specific_date < datetime.now().date():
            tooltip("Easy days can't be applied on past dates.")
            return
        self.specific_dates.append(specific_date)
```

PR #354, opened 2024-01-26
Disperse siblings - Simplify condition, don't limit non-rescheduled cards to maximum ivl

user1823:
- Don't use maximum_interval for cards that are currently scheduled beyond their fuzz limit. Bringing the interval within the max limit is not the function of disperse siblings. The user should use Reschedule for this purpose.
Because we don't want disperse siblings to move cards out of the maximum_interval, I have not changed the code for normal cards.
- The code for dealing with maximum_interval was the same as the one I added to ensure that the cards are not delayed further. So, I just replaced maximum_ivl with current_ivl to remove code complexity.


L-M-Sherlock (approved):
LGTM

PR #357, opened 2024-02-03
Fix/check whether allow to move to past

L-M-Sherlock:
prevent card from moving to past when apply specific easy days

### DIFF
--- a/schedule/reschedule.py
+++ b/schedule/reschedule.py
@@ -17,6 +17,7 @@ class FSRS:
     learned_cnt_perday_from_today: Dict[int, int]
     card: Card
     elapsed_days: int
+    allow_to_past: bool
 
     def __init__(self) -> None:
         self.maximum_interval = 36500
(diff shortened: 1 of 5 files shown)


schedule/reschedule.py
```diff
@@ -72,12 +74,14 @@ def apply_fuzz(self, ivl):
                 return int(self.fuzz_factor * (max_ivl - min_ivl + 1) + min_ivl)
         else:
             min_num_cards = 18446744073709551616
-            best_ivl = (max_ivl + min_ivl) // 2
+            best_ivl = (max_ivl + min_ivl) // 2 if self.allow_to_past else max_ivl
```
user1823, line 77: What does best_ivl actually do? Its exact function is difficult to decipher.

L-M-Sherlock: It sets a default value for best_ivl.

PR #359, opened 2024-02-03
Fix/new fuzz range implemented in Anki

L-M-Sherlock:


user1823:
Don't we need `+ 1` somewhere?

```
/// Return the amount of fuzz to apply to the interval in both directions.
/// Short intervals do not get fuzzed. All other intervals get fuzzed by 1 day
/// plus the number of its days in each defined fuzz range multiplied with the
/// given factor.
```
https://github.com/ankitects/anki/blob/312d396505a4806b76b74dfda6d276ec595e17ed/rslib/src/scheduler/states/fuzz.rs#L102-L105

L-M-Sherlock:
> Don't we need `+ 1` somewhere?

We have:

https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/0ebc299b514093326c741603de9f81bc7672dea5/utils.py#L131

PR #360, opened 2024-02-06
Feat/customize the easy days review ratio

L-M-Sherlock:


user1823:
What about adding a condition so that the random number generator is not called if the ratio is 0 or 1? This will improve performance.

Also, if the ratio is not 0 or 1, what about rescheduling all cards within the next 30 days when Apply Easy Days now is pressed (and not only those that are due on the Easy Days)? After this, the user will not have to reschedule all cards.

user1823:
The patch doesn't work reliably, especially if the user wants to increase the percent after first decreasing it.

What about implementing the following idea?
> if the ratio is not 0 or 1, what about rescheduling all cards due within the next 30 days when Apply Easy Days now is pressed (and not only those that are due on the Easy Days)?

To implement this, I believe that only a small modification is required in the following lines:
https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/aada309c52c7b8c2dc102dc63a363680278ec034/schedule/easy_days.py#L31-L33

user1823:
It seems much better. It's your choice to implement the following; I am fine with merging it in the current state.

If the user has enabled `auto apply easy days`, the current implementation can delay the closing of collection.
So, what if we reschedule the cards due in the next
- 7 days if the auto easy days is enabled and easy days review ratio > 0; and 
- 30 days if any of the above conditions is false.

user1823:
Another idea (this is going to be the last one 😅):

If easy days review ratio > 0, add a "fuzz" around the specific dates in the following code. For example, if the date is 20th Feb, select the cards due from 16th to 24th Feb.

https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/610b0851fd4b843f5633ad08190063105445673d/schedule/easy_days.py#L124-L141

L-M-Sherlock:
I have done my best: [fsrs4anki-helper.ankiaddon.zip](https://github.com/open-spaced-repetition/fsrs4anki-helper/files/14182396/fsrs4anki-helper.ankiaddon.zip)

user1823:
An issue:
I have set ratio = 0.5 and I get the following result:

image

I guess that it is because of the `* 1.25` in the following line:
https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/388d23877812d737bad32aa4f6443b37aace45a3/schedule/reschedule.py#L184

You can merge it after solving the above issue. If it is not a quick solution, you can merge this as it is. The issue can be dealt with in a follow-up. Thank you very much for the time you devoted to this!

I have some nitty-gritty changes in mind (which I have already mentioned before) but they can be implemented in a follow up.

- If the user has enabled auto apply easy days, the current implementation can delay the closing of collection.
 So, what if we reschedule the cards due in the next
   - 7 days if the auto easy days is enabled and easy days review ratio > 0; and
   - 35 days if any of the above conditions is false.
- The "fuzz" around the specific dates should be added only if easy days review ratio > 0. The fuzz doesn't have any benefit when the ratio = 0; it just increases the number of cards that need to be rescheduled.

L-M-Sherlock:
> * After the clicking Apply Easy Days now, there should be a progress bar so that the user knows that the operation is still being executed (rescheduling takes about 8 seconds for me).

I don't know why it doesn't appear. It calls the same function as the `reschedule`.

PR #373, opened 2024-02-28
Burden -> Load & Include buried cards in FSRS stats

user1823:
The word "Burden" is too negative. I have replaced it with a neutral word "Load".

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ The FSRS Stats are based on all cards in your deck or collection (whichever is s
 
 Interpretation:
 
-- Total **burden**, as defined by Piotr Woźniak here: https://supermemo.guru/wiki/Burden
+- **Daily Load** is an estimate of the average review count required per day. It is influenced by the [Burden](https://supermemo.guru/wiki/Burden) statistic in SuperMemo. 
 - The [three component of the memory model](https://supermemo.guru/wiki/Three_component_model_of_memory) used by FSRS:
 	- Average predicted **retention** reflects the percentage of cards that you would recall correctly if you were tested today.
 	- Average (memory) **stability** reflects how fast you forget (forgetting rate). The greater the stability, the slower the forgetting rate.
--- a/stats.py
+++ b/stats.py
@@ -22,7 +22,7 @@ def _lineTbl_now(i):
     return "<table>" + "".join(i) + "</table>"
 
 
-def retention_stability_burden(lim) -> float:
+def retention_stability_load(lim) -> float:
     elapse_stability_ivl_list = mw.col.db.all(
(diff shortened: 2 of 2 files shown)


user1823:
@Expertium, what do you think about it?

aleksejrs:
This introduces a new term for a defined thing:

> Total burden, as defined by Piotr Woźniak here: https://supermemo.guru/wiki/Burden

And uses a word that could be used for something else instead.

user1823:
The current code excludes buried cards from FSRS stats. In the last commit, I have changed it to include them too.

For reference, information about queue is documented here:
https://github.com/ankitects/anki/blob/7784321b90be4ed0b9a7cc157cd88295722cb826/pylib/anki/consts.py#L20

user1823:
> This introduces a new term for a defined thing

Yes, but there is a negative connotation attached with the word "burden". Let us see what Expertium and L-M-Sherlock have to say.

L-M-Sherlock:
I'm OK with that. @Expertium, how about you?

aleksejrs:
A problem with the burden formula is that it does not include the cards currently due/overdue. So a better formula for load could be used. And then the word "Load" will already be used for Burden.

Expertium:
I prefer Burden just because it's how Woz uses the word.
Also, I didn't know that it doesn't include due cards, @L-M-Sherlock can you fix that?

user1823:
> I prefer Burden just because it's how Woz uses the word.

But, I don't think that it is a good reason to reuse that word. FSRS is not meant to be a SuperMemo clone, even though there are several similarities.

By the way, people have been suggesting SuperMemo to change its terminology too. See https://supermemopedia.com/wiki/Do_not_use_the_word_Burden

> Also, I didn't know that it doesn't include due cards

aleksejrs means that if a backlog is present, the burden/load can be smaller than the backlog. However, I think that this doesn't need to be "fixed" because the burden/load is meant to be a measure of the estimated daily review load and the backlog wouldn't exist indefinitely.

Expertium:
> aleksejrs means that if a backlog is present, the burden/load can be smaller than the backlog.

Ah, ok. 

Alright, let's change the name to Daily Load. Daily makes it more clear that it refers to the number of cards per day.

user1823:
> Daily makes it more clear that it refers to the number of cards per day.

I can make that change but isn't it redundant? The load is already expressed in reviews / day.

Expertium:
Never underestimate the user's ability to misunderstand simple things.

user1823:
This PR is ready for final review.

L-M-Sherlock (approved):
LGTM

PR #388, opened 2024-04-15
Feat/knowledge acquisition per hour stat

L-M-Sherlock:


Expertium:
Please add the interpretation of Total Time as well.
**Total Time**: the amount of time spent doing reviews in Anki. This does not include time spent on making/editing cards.

Expertium (approved):

Expertium:
You forgot to change the number of digits after the deicmal point for "Total Time". It should also be 1, not 2.
image

PR #394, opened 2024-04-22
Fix/tooltip not display

L-M-Sherlock:
Patch: [fsrs4anki-helper.ankiaddon.zip](https://github.com/open-spaced-repetition/fsrs4anki-helper/files/15060315/fsrs4anki-helper.ankiaddon.zip)
@user1823, I test it in my device, and it prevents the tooltip from blinking. I guess it could also solve your issue.
@szalejot, this PR also solves your issue.


user1823:
I will test this for a few days and let you know whether the issue was solved or not.

schedule/disperse_siblings.py
```diff
@@ -173,7 +173,7 @@ def disperse_siblings_backgroud(
     undo_entry = mw.col.add_custom_undo_entry("Disperse Siblings")
     mw.taskman.run_on_main(
         lambda: mw.progress.start(
-            label="Siblings Dispersing", max=sibilings_cnt, immediate=False
+            label="Siblings Dispersing", max=sibilings_cnt, immediate=True
```
user1823, line 176:
```suggestion
            label="Dispersing Siblings", max=sibilings_cnt, immediate=True
```

Issue #404, opened 2024-05-08
[Feature Request] Set graph retention rates to 1 if no reviews occured


FSchliephacke:
The retention rate in the retention graph drops to 0 if no reviews were done in a specific range. For me this sometimes happens if I don't have any mature cards on a specific day. Maybe because I reviewed ahead or just couldn't get all my reviews done.

Setting the retention rate to 1 would make the graph easier to read as it wouldn't stretch it out so much.

L-M-Sherlock:
Closed by #405

PR #405, opened 2024-05-08
Feat/Set graph retention rates to 1 if no reviews occured

FSchliephacke:
A simple solution for #404


L-M-Sherlock:
The SQL is hard to read. An alternative method is to add 1 to the denominator and the numerator.

user1823:
Instead of modifying the SQL query, you can probably add a condition just after line 212.

https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/86199dd990d03f664369fa1c1e828a43430a35bd/stats.py#L212

L-M-Sherlock:
Patch: [fsrs4anki-helper.ankiaddon.zip](https://github.com/open-spaced-repetition/fsrs4anki-helper/files/15256529/fsrs4anki-helper.ankiaddon.zip)

This patch may also solve your issue. Could you help me test it?

user1823 (approved):
LGTM

Before:


After:

Issue #406, opened 2024-05-10
[Feature Request] Rework Advance and Postpone


Expertium:
**Is your feature request related to a problem? Please describe.**
Originally, back these features were just introduced, I thought they worked like this:
1) Get R and desired retention
2) Calculate the difference between desired retention and R **after** advancing/postponing
3) Advance/postpone based on that difference

The key point is that we're not comparing desired retention to the current R, before the changes. We are comparing it to R after the changes. For example, if the card is about to be postponed by 1 day, we do not compare R(today) with desired retention, we compare R(today+1) with desired retention. But recently I looked at the code and found that that's not how these things work.

**Describe the solution you'd like**
The method I described above is better than the current implementation. Suppose that we have 2 cards, both with R=90%, and desired retention is 90%. It appears that postponing either of them would have the same effect. However, they have different values of S, and as a result, postponing card 1 would reduce R from 90% to 89%, and postponing card 2 would reduce R from 90% to 80%. The latter produces a larger deviation from optimal scheduling, and hence the first card should be postponed.

user1823:
Firstly, advance is not affected by this issue. In that case, we are actually interested in the retention of the card today because the card is going to become due today.

Now, let's talk about postpone.

> They have different values of S, and as a result, postponing card 1 would reduce R from 90% to 89%, and postponing card 2 would reduce R from 90% to 80%. 

As you can see in the following code, if the R is the same, the card with longer interval is postponed first. So, at a quick glance, the add-on seems to be doing what you have proposed.

https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/fd2ac3e204f62ddc9bd28543423d95682edd96f0/schedule/postpone.py#L73-L74

However, there is a catch.

With the current approach, a card would be postponed from 90% to 80% before another card is postponed from 88% to 86%. If we follow your idea and sort the cards based on their R after postponing, this issue will be solved.

user1823:
I have created a PR: https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/407

Expertium:
> Firstly, advance is not affected by this issue. In that case, we are actually interested in the retention of the card today because the card is going to become due today.

Fair point.

PR #407, opened 2024-05-10
Sort by R after postponing

user1823:
Fixes https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/406

### DIFF
--- a/schedule/postpone.py
+++ b/schedule/postpone.py
@@ -38,8 +38,8 @@ def postpone(did):
             ivl,
             json_extract(data, '$.s'),
             CASE WHEN odid==0
-            THEN {mw.col.sched.today} - (due - ivl)
-            ELSE {mw.col.sched.today} - (odue - ivl)
+            THEN {mw.col.sched.today} - (due - ivl) + ivl * 0.075
+            ELSE {mw.col.sched.today} - (odue - ivl) + ivl * 0.075
             END,
             json_extract(data, '$.dr')
         FROM cards
@@ -55,9 +55,9 @@ def postpone(did):
     # x[1]: did
     # x[2]: interval
     # x[3]: stability
-    # x[4]: elapsed days
+    # x[4]: approx elapsed days after postpone
     # x[5]: desired retention
-    # x[6]: current retention
+    # x[6]: approx retention after postpone
     # x[7]: max interval
     cards = map(
         lambda x: (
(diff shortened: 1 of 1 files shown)


user1823:
The postpone function uses fuzz. So, I have taken the average when estimating the retention to prevent excessive calculations.

schedule/postpone.py
```diff
@@ -38,8 +38,8 @@ def postpone(did):
             ivl,
             json_extract(data, '$.s'),
             CASE WHEN odid==0
-            THEN {mw.col.sched.today} - (due - ivl)
-            ELSE {mw.col.sched.today} - (odue - ivl)
+            THEN {mw.col.sched.today} - (due - ivl) + ivl * 0.075
```
L-M-Sherlock, line 41: What's the purpose of plus ivl * 0.075?

Expertium: I have the same question as well.

user1823: It is used to get an estimate of the elapsed days at the new due date.

The code for the new interval is this:

```py
        delay = elapsed_days - ivl
        new_ivl = min(
            max(1, math.ceil(ivl * (1.05 + 0.05 * random.random())) + delay), max_ivl
        )
```

For the reason explained in https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/407#issuecomment-2104584109, I assume this to be:

```py
        delay = elapsed_days - ivl
        new_ivl = min(
            max(1, math.ceil(ivl * (1.05 + 0.025)) + delay), max_ivl
        )
```

If you put the value of `delay`, you get
```py
        new_ivl = min(
            max(1, math.ceil(ivl * 0.075) + elapsed_days), max_ivl
        )
```

For the purpose of sorting, we can skip `math.ceil` and assume that the value is between `1` and `max_ivl`. So, we get `new_ivl = ivl * 0.075 + elapsed_days`

At due date, `R = power_forgetting_curve(new_ivl, stability)`

Expertium: Oh, interesting. That's a clever way to take fuzz into account.

Expertium (approved):

schedule/postpone.py
user1823: > assume that the value is between 1 and max_ivl

If you don't want to make this assumption, we can change it to something like

```py
THEN min({mw.col.sched.today} - (due - ivl) + ivl * 0.075, max_ivl)
```

Expertium: Yeah, that's better. ivl * 0.075 is more than 1 only if ivl is >13.3(3)

Expertium: Ok

user1823: The value not being > 1 is not actually a problem because it is elapsed_days + ivl * 0.075 and elapsed_days would always be at least equal to 1 (you can't postpone a learning card).

The main concern is that this value can sometimes become greater than the max_ivl but the add-on would not assign that value of interval. So, the retention estimate would become inaccurate.

L-M-Sherlock (approved):
LGTM

Issue #378, opened 2024-03-17
[Feature Request] Store dates for Easy Days


Expertium:
**Is your feature request related to a problem? Please describe.**
Right now, setting a specific date for Easy Days has very little use because the add-on doesn't remember the date.

**Describe the solution you'd like**
Store the dates in add-on config.

L-M-Sherlock:
The current implementation is intended. If we store the specific date, it will be stored in the config.json permanently. And the add-on needs to read them from config when rescheduling. It would make the process more complex.

In addition, specific dates are not periodical like Easy Days of week. It's unnecessary to store them because they will be useless when the dates pass.

Expertium:
But then the user has to re-apply Easy Days to that date.

L-M-Sherlock:
> it makes this feature almost entirely useless

I don't think so. The specific dates are rare.

Expertium:
@user1823 what's your opinion?

user1823:
I think that the feature of Easy Days on specific dates is useless until https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/372 is solved. So, I don't recommend making any changes until we find a solution for https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/372.

> if the best result is achieved by applying Easy Days one day before the specific date, it makes this feature almost entirely useless.

Not entirely useless, but less effective because the reviews can be shifted to the dates ahead but not behind the Easy Day.

peter-moran:
+1 for interest in this. I agree that Easy Days on specific dates does not make sense if it is only a one-time operation (unlike easy days on a specific weekday) since reviews may end up on that day after you set it as an easy day. 

I'd like to use the feature to have some easy days while on vacation, but without it getting memorized if I set the easy day up early it doesn't matter b/c reviews later get scheduled to that day.

L-M-Sherlock:
Have a try: [fsrs4anki-helper.ankiaddon.zip](https://github.com/open-spaced-repetition/fsrs4anki-helper/files/15363349/fsrs4anki-helper.ankiaddon.zip)

PR #416, opened 2024-05-18
Feat/store specific dates for easy days

L-M-Sherlock:


Expertium (approved):
Tbh I doubt that I'll be able to spot a problem even if there is one

L-M-Sherlock:
> Tbh I doubt that I'll be able to spot a problem even if there is one

I uploaded the patch here:

https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/378#issuecomment-2118816043

Could you test it?

Expertium:
I don't use Easy Days personally, so I'd recommend you to ask someone else

PR #422, opened 2024-06-12
Fix/reschedule in n days affects manually rescheduled cards

L-M-Sherlock:


schedule/reschedule.py
```diff
@@ -66,6 +66,7 @@ def set_load_balance(self, did_query=None):
                 f"""SELECT (id/1000-{mw.col.sched.day_cutoff})/86400, count(distinct cid)
                 FROM revlog
                 WHERE ease > 0
+                AND (type < 3 OR factor != 0)
```
user1823, line 69: Is this change required?

If I am not wrong, this code is for calculating the number of reviews done in the past. I don't think that we should exclude the cram revlogs here. Irrespective of whether they are used to reschedule the card or not, they are actual reviews and take up study time just like other reviews do.

L-M-Sherlock: I think cramming is out of plan. Should we consider it in load balance?

user1823: Forget about that for a moment.

Can you first tell me why are calculating the number of reviews done in the past? Load balancing works for future days and not for past days, right?

L-M-Sherlock: Because we need to calculate the size of backlog. If the size of backlog is less than the average, some cards will be rescheduled to today.

user1823: So, do you mean that the past reviews are used to calculate the average? And then, if today's due count is less than the past average, some reviews would be brought forward (aka advanced) from the future.

If so, I agree that the cram reviews shouldn't be used to calculate the average.

By the way, the code is quite complicated. It would be helpful to add some comments explaining the purpose of some of the codes.

L-M-Sherlock: For example, if the size of backlog plus the number of due cards is 100, and the number of due cards tomorrow is 150, some cards due tomorrow will be rescheduled to today.

PR #428, opened 2024-07-02
Fix/don't reschedule to the past when today is easy day

L-M-Sherlock:


user1823:
Why do we need allow_to_past to be `True` in some cases? Shouldn't it always be `False`?

L-M-Sherlock:
> Why do we need `allow_to_past` to be `True` in some cases? Shouldn't it always be `False`?

For example, the user increases the desired retention.

user1823:
But in that case, the following code will make the card due in the past.

https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/4f1d969e691ba2a240cf71d28c552c7662e5e243/schedule/reschedule.py#L91-L93

L-M-Sherlock:
> But in that case, the following code will make the card due in the past.

It's intended. The load balance also tries to balance the backlog and the future workload.

user1823:
I am not saying that those cards should not be scheduled in the past.

I am saying that this code (line 91-93) is enough to ensure that the cards are scheduled in the past when the user increases the desired retention (or updates the parameters). So, you don't need to set `allow_to_past` to True.

L-M-Sherlock:
But it doesn't affect those cards whose fuzz ranges include today.

user1823:
Yes, but what benefit do you get by scheduling a card yesterday (or any other past day) instead of today?

L-M-Sherlock:
It makes the `future due` graph looks smoother because the cards due in yesterday are included in the backlog which is hidden by default.

user1823:
With the current load balancer, the graph will NOT look smooth when there is a backlog.

Reason:
Today's workload includes the backlog but tomorrow's workload doesn't. 
https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/4f1d969e691ba2a240cf71d28c552c7662e5e243/schedule/reschedule.py#L134-L138
https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/4f1d969e691ba2a240cf71d28c552c7662e5e243/schedule/reschedule.py#L61-L65

So, with the current code, after load balancing, today's due count in the Future Due graph will lesser than tomorrow's because today's due count doesn't include the backlog.

If the cards are scheduled today instead of the past, they will be included in today's due count so the graph will be more smooth.

user1823:
Proof:
In the screenshots below, see how the current code makes the today's due count 0 in the Future Due Graph.

Current code:



My code:

L-M-Sherlock:
OK. I will check the PR tonight.

PR #432, opened 2024-07-06
Prevent load balancer from scheduling cards in the past

user1823:
https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/428#issuecomment-2211740347

### DIFF
--- a/schedule/reschedule.py
+++ b/schedule/reschedule.py
@@ -87,7 +87,7 @@ class FSRS:
             # Load balance
             due = self.card.odue if self.card.odid else self.card.due
             last_review = get_last_review_date(self.card)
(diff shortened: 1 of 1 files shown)


user1823:
I think that I must have missed removing some references to the `allow_to_past` variable. Please check that before merging.

user1823:
Also, I need some help. I get this error when using `Reschedule all cards` on my collection with the code in this PR.

```
  File "C:\Users\User\AppData\Roaming\Anki2\addons21\759844606\schedule\reschedule.py", line 117, in apply_fuzz
    for check_ivl in reversed(range(min_ivl, max_ivl + step, step)):
ValueError: range() arg 3 must not be zero

```

L-M-Sherlock (approved):
LGTM

Issue #383, opened 2024-04-07
[BUG] Huge backlog of cards created when I try to apply easy days 


franceskoback:
**Describe the bug**
When I try to follow these instructions (https://www.reddit.com/r/Anki/comments/1attbo1/explaining_fsrs_helper_addon_features/ information ) 
and try to apply easy days, it for some reason just creates a huge backlog of cards that make it so I have over 15000 cards due today instead of 0. 
<img width="753" alt="Screenshot 2024-04-07 at 1 08 04 PM" src="https://github.com/open-spaced-repetition/fsrs4anki-helper/assets/31418742/3d1f3322-65be-4c7b-84ba-bb12e464ec3c">

**To Reproduce**
1. Tools 
2. FRFS anki helper 
3. Less anki on easy days (requires load balancing) --> easy monday
4. Apply easy days now ----> it says "please enable load balance first" so I go to tools and click 
5. Aply load balance when rescheduling. 
6. I then try to click "apply easy days now". It then says "Please either enable FSRS in your deck options, or disable the FSRS helper add-on." so I 
7. go into my Anking deck and press options and then go down to advance and toggle FRFS and press save. It uploads to update all of my many, many cards. Scary. I press save. 
8. I also go into tools and set "easy days review percentage to 70%" !!! so its almost 100%-- shouldn't change the deck that much.  
9. I click "apply easy days now". It says 4380 cards updated in 1.08 seconds. And now I went from having 200 cards left for today to now having over 1800 cards due today.  I have done my anki every day in a row for months and never have more than 300 cards a day. 

**Expected behavior**
I wanted to just have slightly fewer cards on Mondays. 

**Screenshots**

**Environment**
 - Anki version: Version ⁨23.12.1 (1a1d4d54)⁩
Python 3.9.15 Qt 6.5.3 PyQt 6.5.3
 - OS: Mac OS Sonoma 14.2.1 
 - Scheduler version: [e.g. v1.0.0]
 - Optimizer version: [e.g. v1.0.3]
 - Simulator version: [e.g. v1.1.0]

<img width="763" alt="Screenshot 2024-04-07 at 1 16 59 PM" src="https://github.com/open-spaced-repetition/fsrs4anki-helper/assets/31418742/6039fc3b-89fe-4d1a-b050-af54ef7b585c"> 

**Additional context**
Please assist! Would love to have this working, but it's buggy now.

franceskoback:
[...]

PR #433, opened 2024-07-07
Fix/New fuzz range for cards currently scheduled beyond their due range

user1823:
With the current code, the purpose of creating new fuzz ranges is defeated because the card is already given a shorter interval .

For reference, the purpose of creating new fuzz ranges: https://github.com/open-spaced-repetition/fsrs4anki-helper/issues/383#issuecomment-2047985907


L-M-Sherlock (approved):
LGTM

PR #434, opened 2024-07-07
Exclude manual entries from remote reviews

user1823:
Currently, there is no way to distinguish between "Set Due Date" and "Reschedule on change". So, to prevent undoing the "Set Due Date" used on other devices, we need to exclude all manual entries.

When Anki starts using separate revlog entries to distinguish them, we can re-implement the feature of dispersing siblings among cards rescheduled on other devices.


L-M-Sherlock (approved):
LGTM. Thanks!

PR #455, opened 2024-09-13
Feat/remedy hard misuse

L-M-Sherlock:
patch: [fsrs4anki-helper.ankiaddon.zip](https://github.com/user-attachments/files/17006071/fsrs4anki-helper.ankiaddon.zip)

note: the remedy is not undoable with Anki's own framework, so I implement one with my own. For testers, please backup your collection at first.

screenshots:

image<br>

image

image

image

image

image

image

### DIFF
--- a/schedule/remedy.py
+++ b/schedule/remedy.py
@@ -48,7 +48,7 @@ def remedy_hard_misuse(did):
 
     yes = askUser(
         f"""Between {start_date.strftime("%Y-%m-%d")} and {end_date.strftime("%Y-%m-%d")}, {len(revlog_ids)} reviews had a Hard rating.
-These ratings will be replaced with Again. A copy of the unmodified revlogs will be stored in a CSV file in the addon folder to allow undoing the changes.
+These ratings will be replaced with Again. The IDs of these revlogs will be stored in a CSV file in the addon folder to allow undoing the changes.
 Do you want to proceed?
     """
     )
--- a/schedule/remedy.py
+++ b/schedule/remedy.py
@@ -51,7 +51,7 @@ class RemedyDateRangeWidget(QWidget):
         start_date = self.start_date_edit.date().toString("yyyy-MM-dd")
         end_date = self.end_date_edit.date().toString("yyyy-MM-dd")
         yes = askUser(
-            f"Between {start_date} and {end_date}, {len(revlog_ids)} reviews had a Hard rating.\n"
+            f"{len(revlog_ids)} reviews had a Hard rating between {start_date} and {end_date}.\n"
             + "These ratings will be replaced with Again.\n"
             + "The IDs of these revlogs will be stored in a CSV file in the addon folder to allow undoing the changes.\n"
             + "Do you want to proceed?"
--- a/schedule/remedy.py
+++ b/schedule/remedy.py
@@ -43,13 +43,13 @@ def remedy_hard_misuse(did):
     )
 
     if len(revlog_ids) == 0:
-        tooltip("No reviews in the date range")
+        tooltip("There are no reviews with a Hard rating in the selected range of dates.")
         return
 
     yes = askUser(
-        f"""There are {len(revlog_ids)} reviews from {start_date.strftime("%Y-%m-%d")} to {end_date.strftime("%Y-%m-%d")} that are rated as hard.
-Those affected reviews will be rated as again. Are you sure?
-Note: The reviews will be stored in a CSV file in the addon folder for undo.
+        f"""Between {start_date.strftime("%Y-%m-%d")} and {end_date.strftime("%Y-%m-%d")}, {len(revlog_ids)} reviews had a Hard rating.
+These ratings will be replaced with Again. A copy of the unmodified revlogs will be stored in a CSV file in the addon folder to allow undoing the changes.
+Do you want to proceed?
     """
     )
 
--- a/schedule/remedy.py
+++ b/schedule/remedy.py
@@ -51,11 +51,10 @@ class RemedyDateRangeWidget(QWidget):
         start_date = self.start_date_edit.date().toString("yyyy-MM-dd")
         end_date = self.end_date_edit.date().toString("yyyy-MM-dd")
         yes = askUser(
-            f"Between {start_date} and {end_date}, {len(revlog_ids)} reviews had a Hard rating.\n" +
-            "These ratings will be replaced with Again. The requested change will require a one-way sync. " +
-            "If you have made changes on another device, and not synced them to this device yet, please do so before you proceed.\n" +
-            "The IDs of these revlogs will be stored in a CSV file in the addon folder to allow undoing the changes.\n" +
-            "Do you want to proceed?"
+            f"Between {start_date} and {end_date}, {len(revlog_ids)} reviews had a Hard rating.\n"
+            + "These ratings will be replaced with Again.\n"
+            + "The IDs of these revlogs will be stored in a CSV file in the addon folder to allow undoing the changes.\n"
+            + "Do you want to proceed?"
         )
 
         if not yes:
@@ -82,6 +81,9 @@ class RemedyDateRangeWidget(QWidget):
 
 
 def remedy_hard_misuse(did):
+    if not ask_one_way_sync():
+        return
+
     mw.date_range_widget = RemedyDateRangeWidget()
     mw.date_range_widget.show()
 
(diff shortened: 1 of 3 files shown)


Expertium:
[The number of people who misuse Hard is guaranteed to be greater than the number of people who misuse Hard AND use the FSRS Helper add-on and will notice this feature and use it](https://en.wikipedia.org/wiki/Conjunction_fallacy). And it doesn't address the underlying problem, which can only be solved by better UI design and tooltips.

Expertium:
Just to be clear, I'm not saying that this should be canceled. This feature deserves to exist, I just don't think that this will be the best solution to the problem, or even the second best solution, or even make much of a difference.

schedule/remedy.py
```diff
@@ -0,0 +1,98 @@
...
+        f"""SELECT id
+        FROM revlog
+        WHERE ease = 2
+        AND id > {start_date.timestamp() * 1000}
+        AND id < {end_date.timestamp() * 1000}
+        """
+    )
+
+    if len(revlog_ids) == 0:
+        tooltip("No reviews in the date range")
```
user1823, line 46:
```suggestion
        tooltip("There are no reviews with a Hard rating in the selected range of dates.")
```

schedule/remedy.py
```diff
@@ -0,0 +1,98 @@
...
+        return
+
+    yes = askUser(
+        f"""There are {len(revlog_ids)} reviews from {start_date.strftime("%Y-%m-%d")} to {end_date.strftime("%Y-%m-%d")} that are rated as hard.
```
user1823, line 50:
```suggestion
        f"""Between {start_date.strftime("%Y-%m-%d")} and {end_date.strftime("%Y-%m-%d")}, {len(revlog_ids)} reviews had a Hard rating.
```

schedule/remedy.py
```diff
@@ -0,0 +1,98 @@
...
+Those affected reviews will be rated as again. Are you sure?
```
user1823, line 51:
```suggestion
These ratings will be replaced with Again. A copy of the unmodified revlogs will be stored in a CSV file in the addon folder to allow undoing the changes.
```

schedule/remedy.py
```diff
@@ -0,0 +1,98 @@
...
+Note: The reviews will be stored in a CSV file in the addon folder for undo.
```
user1823, line 52:
```suggestion
Do you want to proceed?
```

user1823:
> image

Can you replace it with an input box like the one in Easy Days on Specific Dates?

L-M-Sherlock:
> Can you replace it with an input box like the one in Easy Days on Specific Dates?

OK. I will improve it tomorrow.

user1823:
This will also require a one-way sync. So, we need to add this message.

```
The requested change will require a one-way sync. If you have made changes on another device, and not synced them to this device yet, please do so before you proceed.
```

I am not sure if add-ons can use Anki's strings. If they can, you can just add `tr.deckConfigWillRequireFullSync()`

schedule/remedy.py
```diff
@@ -0,0 +1,98 @@
...
+    if not yes:
+        return
+
+    mw.col.db.execute(
+        f"""UPDATE revlog
+        SET ease = 1
+        WHERE id IN {ids2str(revlog_ids)}
+        """
+    )
+
```
user1823, line 65:
```suggestion
    mw.set_schema_modified()

```

Please check if this works correctly or needs some modification (like using self.set_schema_modified() instead).

schedule/remedy.py
```diff
@@ -0,0 +1,98 @@
...
+    with open(revlog_id_csv, "r") as f:
+        revlog_ids = list(map(int, f.read().splitlines()))
+
+    mw.col.db.execute(
+        f"""UPDATE revlog
+        SET ease = 2
+        WHERE id IN {ids2str(revlog_ids)}
+        """
+    )
+
```
user1823, line 95:
```suggestion
    mw.set_schema_modified()

```

Please check if this works correctly or needs some modification (like using self.set_schema_modified() instead).

schedule/remedy.py
```diff
@@ -0,0 +1,98 @@
...
+    if len(revlog_ids) == 0:
+        tooltip("There are no reviews with a Hard rating in the selected range of dates.")
+        return
+
+    yes = askUser(
+        f"""Between {start_date.strftime("%Y-%m-%d")} and {end_date.strftime("%Y-%m-%d")}, {len(revlog_ids)} reviews had a Hard rating.
+These ratings will be replaced with Again. A copy of the unmodified revlogs will be stored in a CSV file in the addon folder to allow undoing the changes.
```
user1823, line 51:
```suggestion
These ratings will be replaced with Again. The IDs of these revlogs will be stored in a CSV file in the addon folder to allow undoing the changes.
```
Previously, I had a wrong idea of what was stored in the CSV file.

schedule/remedy.py
L-M-Sherlock: we don't need to do the one-way sync. Only modified revlogs need sync.

user1823: I tried using the latest version (https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/455/commits/c29752cec806f35e49166341851412f4cbd7a5f4) and the modified revlogs weren't synced to the other device.

As far as I know, revlogs once created can't be modified by sync unless a one-way sync is performed.

L-M-Sherlock: Thanks for the report. I forget to check it in my phone. Seems the modified revlogs have been uploaded, but the other device doesn't update it during sync. So I have to use `set_schema_modified()`.

user1823:
Can we make the one-way sync warning pop up immediately after the user chooses `Remedy` or `Undo` from the add-on menu?

L-M-Sherlock:
I haven't found the API to trigger one-way sync.

user1823:
> I haven't found the API to trigger one-way sync.

Sorry, I didn't understand. Maybe you misunderstood my comment. I meant adding something like 

```py
showWarning(tr.deckConfigWillRequireFullSync())
```
between these two lines:
https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/f902af4df00f42c53f55e28f730be773a7a650d6/schedule/remedy.py#L84-L85

and between these two lines:
https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/f902af4df00f42c53f55e28f730be773a7a650d6/schedule/remedy.py#L89-L90

L-M-Sherlock:
> ```python
> showWarning(tr.deckConfigWillRequireFullSync())
> ```

Anki add-on has no access to the translation strings of Anki.

L-M-Sherlock:
How about that?

image

image

user1823:
The text in the remedy option is too long. Users are likely to ignore such walls of text.

A better way would be to add the one-way sync warning before the user reaches the dialog for choosing the start and end dates.

user1823:
> Anki add-on has no access to the translation strings of Anki.

Well, I tested this and addons CAN access the translated strings.

Adding this to the add-on worked:
```py
from aqt.utils import tr
showWarning(tr.deck_config_will_require_full_sync())
```

However, askUser is probably better than showWarning because it also allows the user to go back and perform a sync before continuing.

Moreover, in askUser, we will also add "Do you want to proceed?" and having some text translated and some untranslated is not a good UX imo.

So, I think that just copying the text into the add-on is the better option.

As a side note, using just tr.deck_config_will_require_full_sync() in Python causes unexpected line breaks to appear. Dae has shared the solution for the issue here: https://github.com/ankidroid/Anki-Android/issues/13040#issuecomment-1365722828

An example of this in use:
```py
            showWarning(with_collapsed_whitespace(tr.errors_100_tags_max()))
```
https://github.com/ankitects/anki/blob/73c97de5d022abcec0ab1fb70921c6bf813cfa28/qt/aqt/taglimit.py#L89

L-M-Sherlock:
So you mean just to keep the current design, right?

schedule/remedy.py
```diff
@@ -51,11 +51,10 @@ def remedy_hard_misuse(self):
         start_date = self.start_date_edit.date().toString("yyyy-MM-dd")
         end_date = self.end_date_edit.date().toString("yyyy-MM-dd")
         yes = askUser(
-            f"Between {start_date} and {end_date}, {len(revlog_ids)} reviews had a Hard rating.\n" +
-            "These ratings will be replaced with Again. The requested change will require a one-way sync. " +
-            "If you have made changes on another device, and not synced them to this device yet, please do so before you proceed.\n" +
-            "The IDs of these revlogs will be stored in a CSV file in the addon folder to allow undoing the changes.\n" +
-            "Do you want to proceed?"
+            f"Between {start_date} and {end_date}, {len(revlog_ids)} reviews had a Hard rating.\n"
```
user1823, line 54:
```suggestion
            f"{len(revlog_ids)} reviews had a Hard rating between {start_date} and {end_date}.\n"
```

user1823:
> So you mean just to keep the current design, right?

Yes. Now, I have just two suggestions:
1. The wording suggestion posted just now.
2. Extract the `def ask_one_way_sync()` into some other file (like utils.py) because we might want to use it in some other context later.

After these two changes are done, treat it as approved.

PR #461, opened 2024-09-19
Refactor/load balance & easy days

L-M-Sherlock:
The `load balance` and `easy days` features are refactored in the way discussed in:
- https://github.com/open-spaced-repetition/load-balance-simulator/issues/1#issuecomment-2361674400
- https://github.com/open-spaced-repetition/easy-days-simulator/issues/1#issuecomment-2294822529
- https://github.com/ankitects/anki/issues/3116#issuecomment-2295243077

Patch: [fsrs4anki-helper.ankiaddon.zip](https://github.com/user-attachments/files/17069049/fsrs4anki-helper.ankiaddon.zip)

Preview:

image

image

### DIFF
--- a/schedule/easy_days.py
+++ b/schedule/easy_days.py
@@ -25,21 +25,30 @@ def easy_days(did):
     if not config.load_balance:
         tooltip("Please enable load balance first")
         return
-    if len(config.easy_days) == 0 and len(config.easy_dates) == 0:
+    if (
+        all([r == 1 for r in config.easy_days_review_ratio_list])
+        and len(config.easy_dates) == 0
+    ):
         tooltip("Please select easy days or specific dates first")
         return
     today = mw.col.sched.today
     due_days = []
-    if config.easy_days_review_ratio > 0 and config.auto_easy_days:
+    if (
+        any([r > 0 for r in config.easy_days_review_ratio_list])
+        and config.auto_easy_days
+    ):
         reschedule_range = 7
     else:
         reschedule_range = 35
     for day_offset in range(reschedule_range):
         if (
-            config.easy_days_review_ratio
-            > 0  # if ratio > 0, reschedule all cards due in the reschedule_range
-            or (sched_current_date() + timedelta(days=day_offset)).weekday()
-            in config.easy_days  # if ratio = 0, reschedule only those due on Easy days
+            any(
+                [r > 0 for r in config.easy_days_review_ratio_list]
+            )  # if ratio > 0, reschedule all cards due in the reschedule_range
+            or config.easy_days_review_ratio_list[
+                (sched_current_date() + timedelta(days=day_offset)).weekday()
+            ]
+            < 1  # if ratio = 0, reschedule only those due on Easy days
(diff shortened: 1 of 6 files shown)


Expertium:
I don't see why you are doing this in the first place. If Easy Days and Load Balance will work the same way both natively and in the add-on, what reason would there be to keep both? If one offers different pros and cons, sure. But if the add-on versions and the native versions are the same, why keep duplicate functionality?

L-M-Sherlock:
Firstly, I don't know when the Easy Days will be implemented natively. Secondly, implementing it in the add-on could help us the method easily.

L-M-Sherlock:
By the way, it really helped me improve the method.

Before https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/461/commits/43a61aa4e747bc39df9a107c1065ec25a285579e:

image

After https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/461/commits/43a61aa4e747bc39df9a107c1065ec25a285579e:

image

Expertium (approved):

brishtibheja:
@expertium a lot of users don't update and intentionally/unintentionally keep themselves on old versions. So it might be useful for them.

Expertium:
I would replace "Reduce" with "Reduced" and "Minimize" with "Minimimum".

schedule/easy_days.py
```diff
@@ -23,21 +25,20 @@ def easy_days(did):
...
+    for day_offset in range(35):
         if (
-            config.easy_days_review_ratio
-            > 0  # if ratio > 0, reschedule all cards due in the reschedule_range
-            or (sched_current_date() + timedelta(days=day_offset)).weekday()
-            in config.easy_days  # if ratio = 0, reschedule only those due on Easy days
+            config.easy_days_review_ratio_list[
+                (sched_current_date() + timedelta(days=day_offset)).weekday()
+            ]
+            < 1  # if ratio = 0, reschedule only those due on Easy days
```
user1823, line 41: What does it do?
I mean, when would easy_days_review_ratio_list be < 1?

schedule/easy_days.py
```diff
@@ -23,21 +25,20 @@ def easy_days(did):
...
         tooltip("Please select easy days or specific dates first")
         return
     today = mw.col.sched.today
     due_days = []
-    if config.easy_days_review_ratio > 0 and config.auto_easy_days:
-        reschedule_range = 7
-    else:
-        reschedule_range = 35
-    for day_offset in range(reschedule_range):
+    for day_offset in range(35):
```
user1823, line 36: Why was the code removed here? The previous code was aimed at improving performance when the user is using "Auto apply easy days on closing collection".

L-M-Sherlock: Because `config.easy_days_review_ratio` has been removed, which slowed down the performance in the previous code. After refactor, the performance is improved, so I think it's OK to remove this condition.

schedule/easy_days.py
L-M-Sherlock: It's less than one in `Reduced` and `Minimum` modes.

PR #462, opened 2024-09-20
Feat/reschedule cards only when the odds is different by 25%

### LINKED ISSUE #419: [Feature Request] Reschedule cards only if the new interval is different by >x%
**Is your feature request related to a problem? Please describe.**
This is a continuation of the Discord discussion. Many users avoid rescheduling because it greatly affects their workload even if parameters changed very slightly. Solution: only reschedule cards if the new interval is x% longer/shorter than before rescheduling. Once we test this, it can be implemented natively.

**Describe the solution you'd like**
Add a configurable threshold from 0% to 30%, and I'll test how much it affects the number of due cards when I reschedule.

L-M-Sherlock:
close #419 

patch: [fsrs4anki-helper.ankiaddon.zip](https://github.com/user-attachments/files/17072521/fsrs4anki-helper.ankiaddon.zip)

However, it would break load balance & easy days.

### DIFF
--- a/schedule/reschedule.py
+++ b/schedule/reschedule.py
@@ -375,7 +375,7 @@ def reschedule_card(cid, fsrs: FSRS, recompute=False):
         if card.ivl <= adjusted_ivl:
             return None
 
-        incresed_odds = 1.25 * odds
+        increased_odds = 1.25 * odds
         fsrs.desired_retention = incresed_odds / (incresed_odds + 1)
         adjusted_ivl = fsrs.next_interval(s)
 
--- a/schedule/reschedule.py
+++ b/schedule/reschedule.py
@@ -364,6 +364,24 @@ def reschedule_card(cid, fsrs: FSRS, recompute=False):
         fsrs.set_card(card)
         fsrs.set_fuzz_factor(cid, card.reps)
         new_ivl = fsrs.next_interval(s)
+
+        dr = fsrs.desired_retention
+        odds = dr / (1 - dr)
+
+        reduced_odds = 0.75 * odds
+        fsrs.desired_retention = reduced_odds / (reduced_odds + 1)
+        adjusted_ivl = fsrs.next_interval(s)
+
+        if card.ivl <= adjusted_ivl:
+            return None
+        
+        incresed_odds = 1.25 * odds
+        fsrs.desired_retention = incresed_odds / (incresed_odds + 1)
+        adjusted_ivl = fsrs.next_interval(s)
+
+        if card.ivl >= adjusted_ivl:
+            return None
+
         due_before = card.odue if card.odid else card.due
         card = update_card_due_ivl(card, new_ivl)
         due_after = card.odue if card.odid else card.due
--- a/schedule/reschedule.py
+++ b/schedule/reschedule.py
@@ -160,7 +160,7 @@ class FSRS:
             )
             return best_ivl
 
-    def next_interval(self, stability):
+    def fuzzed_next_interval(self, stability):
         new_interval = next_interval(stability, self.desired_retention)
         return self.apply_fuzz(new_interval)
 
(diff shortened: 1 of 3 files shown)


Expertium:
I'll copy what I said in the related issue

If you want me to test different values, you can make `reduced_odds = (1-x)*odds` and `increased_odds = (1+x)*odds`, where x can be configured in add-on config.

schedule/reschedule.py
```diff
@@ -364,6 +364,24 @@ def reschedule_card(cid, fsrs: FSRS, recompute=False):
...
+
+        reduced_odds = 0.75 * odds
+        fsrs.desired_retention = reduced_odds / (reduced_odds + 1)
+        adjusted_ivl = fsrs.next_interval(s)
+
+        if card.ivl <= adjusted_ivl:
+            return None
+        
+        incresed_odds = 1.25 * odds
+        fsrs.desired_retention = incresed_odds / (incresed_odds + 1)
```
Expertium, line 379: incresed -> incre**a**sed

Expertium:
@user1823

>I think that this feature deserves it own entry.

I disagree. First of all, if there are two rescheduling features, the user will be like "So this is rescheduling and this is also rescheduling...what's the difference? Which one should I use?". Second of all, ideally, this new change should make rescheduling better (in the sense that the backlog will be smaller), and if it's better, it should supercede the current method.

schedule/reschedule.py
```diff
@@ -364,6 +364,24 @@ def reschedule_card(cid, fsrs: FSRS, recompute=False):
...
+        odds = dr / (1 - dr)
+
+        reduced_odds = 0.75 * odds
+        fsrs.desired_retention = reduced_odds / (reduced_odds + 1)
+        adjusted_ivl = fsrs.next_interval(s)
+
+        if card.ivl <= adjusted_ivl:
+            return None
+        
+        incresed_odds = 1.25 * odds
```
L-M-Sherlock, line 378:
```suggestion
        increased_odds = 1.25 * odds
```

L-M-Sherlock:
[fsrs4anki-helper.ankiaddon.zip](https://github.com/user-attachments/files/17083845/fsrs4anki-helper.ankiaddon.zip)

image

Expertium:
Alright, I tested the backlog that I get with different thresholds on one of my decks
0% - 1024 due cards
5% - 1030
10% - 1035
15% - 1018
20% - 960
25% - 1017
30% - 889
35% - 834

I don't know what's going on. For some reason not only is the backlog not decreasing much, but for some values it actually **increases** compared to when the threshold is 0%.

@user1823 give it a try

user1823:
> I don't know what's going on. 

Did you disable `Load Balance` and `Disperse Siblings after rescheduling`? Even though the user might want to use these options, we need to disable these during testing so that we can see the trend.

## My results

Before rescheduling: 101 cards due

After:
0% - 2184
10% - 2074
15% - 1751
20% - 1340
25% - 950
30% - 606
40% - 222

user1823:
Some suggestions:
- When "Disperse siblings after rescheduling" is enabled, reschedule ONLY the sibling whose current interval deviates the most from the ideal one. The disperse function that runs automatically will handle the rest. The reason for this request is to prevent >1 sibling from becoming due after rescheduling. I guess that implementing this would be slightly hard but I think that its worth it.
- Before `if card.ivl >= adjusted_ivl_lower and card.ivl <= adjusted_ivl_upper:`, add `fsrs.desired_retention = dr`. Currently, this will be a non-functional change. But, it can prevent many bugs in the future.
- Rename `reduced_odds` and `increased_odds` to `odds_lower` and `odds_upper` respectively.

Expertium:
> Did you disable `Load Balance` and `Disperse Siblings after rescheduling`? Even though the user might want to use these options, we need to disable these during testing so that we can see the trend.

Nope, my bad.

user1823:
I also tried with `Disperse Siblings after rescheduling` enabled. Here are the results:

Before rescheduling: 101 cards due

After:
0% - 2075
10% - 1987
15% - 1689
20% - 1313
25% - 963
30% - 653
40% - 292

The results show the same trend as above.

user1823:
> Based on your results, 25% seems good. It reduces the backlog in half.

Yes. It will reduce further after the first suggestion in https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/462#issuecomment-2365239994 is implemented.

Expertium:
Alright, I turned off Load Balancing.
0%	1320 due cards
5%	1318
10%	1314
15%	1300
20%	1264
25%	1175
30%	1094
35%	1044
40%	958

While the change isn't as big as in your case, it's clear that it does help at least somewhat.

user1823:
By the way, as L-M-Sherlock said, in the current form, this feature breaks load Balancing and Easy Days because the add-on won't reschedule the cards whose intervals had to change by a small amount.

We can easily fix Easy Days by adding `and not self.apply_easy_days` to `if card.ivl >= adjusted_ivl_lower and card.ivl <= adjusted_ivl_upper:`.

However, fixing load balancing requires either
- creating a new option for Load Balancing
- creating a new option for this feature.

IMO, the second choice is better because it will also allow the user to completely reschedule their cards even if they don't want to use load balancing.

schedule/reschedule.py
```diff
@@ -364,6 +367,21 @@ def reschedule_card(cid, fsrs: FSRS, recompute=False):
         fsrs.set_card(card)
         fsrs.set_fuzz_factor(cid, card.reps)
         new_ivl = fsrs.next_interval(s)
+
+        dr = fsrs.desired_retention
+        odds = dr / (1 - dr)
+
+        reduced_odds = (1 - fsrs.reschedule_threshold) * odds
+        fsrs.desired_retention = reduced_odds / (reduced_odds + 1)
+        adjusted_ivl_upper = fsrs.next_interval(s)
```
user1823, line 376: Is fuzz applied when calculating `adjusted_ivl_upper` and `adjusted_ivl_lower`? Ideally, fuzz should not be applied.

I am asking this because apparently there are two different functions with the same name.

https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/8c183e9fb53c4812831914bd8ba84905823fddfd/schedule/reschedule.py#L163-L165

PR #469, opened 2024-09-26
Patch for #462

user1823:
After these changes, https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/462 should be ready for merge. The remaining issues can be tackled in another PR.

Remaining requests/issues:
- When "Disperse siblings after rescheduling" is enabled, reschedule ONLY the sibling whose current interval deviates the most from the ideal one. The disperse function that runs automatically will handle the rest. The reason for this request is to prevent >1 sibling from becoming due after rescheduling. I guess that implementing this would be slightly hard but I think that it's worth it.
- Fixing load balancing requires either creating a new option for Load Balancing or creating a new option for this feature. IMO, the second choice is better because it will also allow the user to completely reschedule their cards even if they don't want to use load balancing.

### DIFF
--- a/config.json
+++ b/config.json
@@ -8,7 +8,7 @@
     "auto_disperse_when_review": false,
     "auto_disperse_after_reschedule": false,
     "mature_ivl": 21,
-    "reschedule_threshold": 0.25,
+    "reschedule_threshold": 0,
     "debug_notify": false,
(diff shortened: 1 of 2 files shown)


L-M-Sherlock (approved):
LGTM

PR #471, opened 2024-09-29
Fix/filter out manual and cram reviews in true retention & use _revlogLimit()

L-M-Sherlock:
fix: 
- https://github.com/ankitects/anki/issues/3453

### DIFF
--- a/stats.py
+++ b/stats.py
@@ -86,7 +86,6 @@ def retention_stability_load(lim) -> tuple:
     )
     """
     )
-    print(time_sum)
     return (
         recall_sum / card_cnt,
         stability_sum / card_cnt,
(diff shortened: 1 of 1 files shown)


user1823:
I don't know why but this doesn't seem to have any effect.

user1823:
Well, that's because the code for true retention already filters the revlogs with `ease = 0` and I don't have any revlogs with `type = 3 and factor = 0`.

L-M-Sherlock:
> Well, that's because the code for true retention already filters the revlogs with `ease = 0`

The previous version doesn't.

user1823:
> The previous version doesn't.

It does filter them in the following lines:
https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/a459532f73c9772398503491035b7bfbc5f78bdf/stats.py#L389-L392

Anyway, the changes LGTM. They don't solve the issue reported by me, though.

L-M-Sherlock:
> It does filter them in the following lines:

Nope. It just treats them as failed (0). That causes the previous discrepancy.

user1823:
Well, these lines include `ease = 1` and `ease > 1`. So, I don't know how the add-on is counting the revlogs with ease = 0.

I think that the discrepancy is corrected because the latest commit in Anki uses the correct code to remove the crammed cards.

L-M-Sherlock:
Fine. The bug has been fixed. It's a good news.

PR #476, opened 2024-10-09
Fix/next_interval() missing 1 required positional argument: 'r'

user1823:


L-M-Sherlock:
Why not use `self.fuzzed_next_interval`?

user1823:
Because it also applies fuzz, which means that the max and min odds will not be exactly the same as calculated.

L-M-Sherlock (approved):
LGTM

PR #485, opened 2024-10-29
Fix/user should has one normal day at least when setting easy days

Expertium:
I don't know if this works properly, and there is probably a more efficient way of doing it (with a single loop), but I absolutely will not let this go unfixed. LMSherlock, if you can do better than what I did in this PR - please do so.

### DIFF
--- a/schedule/easy_days.py
+++ b/schedule/easy_days.py
@@ -251,19 +251,15 @@ class EasyDaysReviewRatioSelector(QWidget):
         self.resize(400, 250)
 
     def save_settings(self):
-        # first loop, for checking whether the user has set all days to "Reduced" or as "Minimum"
-        reduced_days = 0
-        minimum_days = 0
-        for day in self.weekdays:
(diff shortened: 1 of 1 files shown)


L-M-Sherlock:
What if the user has 6 reduced days and 1 minimum day? It's equal to 6 normal days and 1 minimum day. Should we warn the user in this case?

Expertium:
Good question. I'd say yes. I think we should allow the user to do that, but display a warning "Your current configuration is equivalent to ..."

L-M-Sherlock:
So the condition is the user should have one normal day at least.

Expertium:
The condition for not showing any warnings? Yes.

L-M-Sherlock (approved):
LGTM

Expertium:
Yeah, I think that's good. And please do this for the native implementation too, thank you.

L-M-Sherlock:
The solution provided by sonnet doesn't work. I don't know how to implement it in Anki. The reactive frontend framework is confusing me. Please ask for other developer.

PR #488, opened 2024-11-10
Feat/clear manual rescheduling logs

L-M-Sherlock:
Preview:

image

Patch:

[fsrs4anki-helper.ankiaddon.zip](https://github.com/user-attachments/files/17690719/fsrs4anki-helper.ankiaddon.zip)


user1823:
Like https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/455, this should force a one-way sync.

Luc-Mcgrady:
image
It might be possible to separate from other schedule changes based on it having a usn of not -1.

L-M-Sherlock:
@user1823 nice catch. The last 90ed304 comment should solve it.

L-M-Sherlock:
@user1823 could you help me improve the wording?

user1823:
Can you first explain what is the actual purpose of this feature? Is it aimed at removing just the "Reschedule cards on change" entries with the other entries being removed only because we can't keep them?

L-M-Sherlock:
Yes, it aims at removing the useless entires generated by "Reschedule cards on change". In extreme case, the percentage of this kind of useless entires will reach 25%.

@Expertium, I hope you can elaborate it: https://discord.com/channels/368267295601983490/1282005522513530952/1304016208080666634

user1823:
In Anki 24.11+, we can distinguish between `Reschedule cards on change` and `Set due date` entries (only those entries which were created after the update). However, if we still have to use the same SQL query because the older entries are ambiguous, I am wondering what was the benefit of that change.

user1823:
Probably we need to make it more configurable. I am thinking something along these lines:
- In Anki 24.11+, this feature should, by default, only remove the entries with `type = 5`.
- The user can choose to remove the older entries with the warning that it will also remove Set Due Date entries and a message that they need to select this option only once (because the newer Reschedule on change entries can be removed with the default query).

@Expertium, WDYT? If you agree, can you help in designing the UI for this?

Luc-Mcgrady:
What I've always thought would be a good solution for this would be to delete any manual review that comes directly before another manual review. That way the set due dates/reschedules that were never reviewed would be cleared while still leaving any reschedules/manual sets that actually matter.

Example:
image

L-M-Sherlock:
> What I've always thought would be a good solution for this would be to delete any manual review that comes directly before another manual review.

Could you help write the SQL? Or implement it in another way?

Luc-Mcgrady:
```sql
DELETE FROM revlog as cur
WHERE cur.type >= 4
AND cur.factor <> 0
AND (
	SELECT type
	FROM revlog
	WHERE id = (
		SELECT min(id)
		FROM revlog
		WHERE id > cur.id
		AND cid == cur.cid
	)
) >= 4
``` 
This should work

__init__.py
```diff
@@ -179,6 +179,10 @@ def reschedule_recent(did):
 
 menu_reset = build_action(clear_custom_data, "Clear custom data in all cards")
 
+menu_clear_manual_rescheduling = build_action(
+    clear_manual_rescheduling, "Clear manual rescheduling in all cards"
```
user1823, line 183:
```suggestion
    clear_manual_rescheduling, "Delete redundant manual revlog entries"
```

schedule/reset.py
```diff
@@ -33,3 +33,39 @@ def clear_custom_data(did):
...
     tooltip(f"""{cnt} cards cleared in {time.time() - start_time:.2f} seconds.""")
     mw.reset()
+
+
+def clear_manual_rescheduling(did):
+    if not ask_one_way_sync():
+        return
+
+    if not askUser(
+        "<p>Delete any manual review that comes directly before another manual review?</p>"
```
user1823, line 43: I suggest replacing all of this text by:

> This option removes revlog entries generated by "Set due date" and "Reschedule cards on change" that immediately precede another manual or reschedule entry. 
>
> This action cannot be undone. Do you want to continue?

schedule/reset.py
```diff
@@ -33,3 +33,39 @@ def clear_custom_data(did):
...
+                SELECT min(id)
+                FROM revlog
+                WHERE id > cur.id
+                AND cid == cur.cid
+            )
+        ) >= 4
+    """
+    )
+    col_set_modified()
+    tooltip("Review logs cleared.")
```
user1823, line 70: It would be better if we can also show the number of revlogs deleted.

user1823 (changes requested):

L-M-Sherlock:
Done:

[fsrs4anki-helper.ankiaddon.zip](https://github.com/user-attachments/files/17741813/fsrs4anki-helper.ankiaddon.zip)

khonkhortisan:
Delete redundant manual revlog entries → force full sync? [yes] → remove revlog entries generated by "Set due date"? [no] → normal sync [y]. The yes/no dialogs should be in the other order, so that I'm locked into full syncing immediately after the full sync dialog.

schedule/reset.py
```diff
@@ -33,3 +34,39 @@ def clear_custom_data(did):
...
+                WHERE id > cur.id
+                AND cid == cur.cid
+            )
+        ) >= 4
+    """
+    )
+    cnt = len(revlog_ids)
+    mw.col.db.execute(f"DELETE FROM revlog WHERE id IN {ids2str(revlog_ids)}")
+    col_set_modified()
+    tooltip(f"{cnt} review logs cleared.")
```
user1823, line 71:
```suggestion
    tooltip(f"{cnt} review logs deleted.")
```

user1823 (commented):
- One minor wording suggestion
- I agree with the suggestion by khonkhortisan. The order of the two yes/no dialogs should be interchanged.
- Because the working of the option has changed, the name of the option (clear_manual_rescheduling) looks slightly incorrect.

After these are dealt with, I think that this PR can be merged.

L-M-Sherlock:
Sorry, I'm confused. I think `force full sync` is necessary.

user1823:
Yes, it is necessary.

But, move the full sync dialog AFTER you tell the user what this option does.

So,  `if not ask_one_way_sync(): return` comes just before `revlog_ids = mw.col.db.list(`

L-M-Sherlock:
cc

[fsrs4anki-helper.ankiaddon.zip](https://github.com/user-attachments/files/17749030/fsrs4anki-helper.ankiaddon.zip)

khonkhortisan:
Delete redundant manual revlog entries → force full sync? [yes] → remove revlog entries generated by "Set due date"? [no] → normal sync [y]. The yes/no dialogs should be in the other order, so that I'm locked into doing a full sync immediately after accepting.

PR #502, opened 2024-11-30
Feat/preset level easy days

L-M-Sherlock:
Because the smart fuzz has been built-in, so the load balance option has been removed. The load balance and easy days are always applied when rescheduling via the helper add-on.

Patch: [fsrs4anki-helper.ankiaddon.zip](https://github.com/user-attachments/files/17967907/fsrs4anki-helper.ankiaddon.zip)

After this PR, I will release a new branch in Ankiweb.

image

The new branch will only support Anki 24.11+ versions.

### DIFF
--- a/schedule/easy_days.py
+++ b/schedule/easy_days.py
@@ -22,27 +22,8 @@ from aqt.gui_hooks import profile_will_close
 def easy_days(did):
     config = Config()
     config.load()
-    if not config.load_balance:
-        tooltip("Please enable load balance first")
-        return
-    if (
-        all([r == 1 for r in config.easy_days_review_ratio_list])
-        and len(config.easy_dates) == 0
-    ):
-        tooltip("Please select easy days or specific dates first")
-        return
     today = mw.col.sched.today
-    due_days = []
-    for day_offset in range(35):
-        if (
-            config.easy_days_review_ratio_list[
-                (sched_current_date() + timedelta(days=day_offset)).weekday()
-            ]
-            < 1
-            or (sched_current_date() + timedelta(days=day_offset)).strftime("%Y-%m-%d")
-            in config.easy_dates
-        ):
-            due_days.append(today + day_offset)
+    due_days = [today + day_offset for day_offset in range(35)]
 
     # find cards that are due in easy days in the next 35 days
     due_in_easy_days_cids = mw.col.db.list(
@@ -62,7 +43,6 @@ def easy_days(did):
         recent=False,
         filter_flag=True,
         filtered_cids=set(due_in_easy_days_cids),
-        apply_easy_days=True,
     )
     if fut:
         return fut.result()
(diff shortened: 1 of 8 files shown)


schedule/reschedule.py
```diff
@@ -171,43 +174,37 @@ def apply_fuzz(self, ivl):
...
-            best_ivl = self.load_balance(
-                possible_intervals,
-                review_cnts,
-                last_review,
+
+        # Load balance
+        due = self.card.odue if self.card.odid else self.card.due
+        last_review = get_last_review_date(self.card)
+
+        if due > last_review + max_ivl + 2:
```
user1823, line 182: We want this to apply only when the user presses "Apply easy days now" or uses the easy days on specific days feature.

Otherwise, the user won't be able to correct the due date of cards whose due date is larger than the max limit of fuzz range.

schedule/reschedule.py
```diff
@@ -415,7 +405,7 @@ def reschedule_card(cid, fsrs: FSRS, recompute=False):
         fsrs.set_fuzz_factor(cid, card.reps)
         new_ivl = fsrs.fuzzed_next_interval(s)
 
-        if fsrs.reschedule_threshold > 0 and not fsrs.apply_easy_days:
+        if fsrs.reschedule_threshold > 0:
```
user1823, line 408: This should be reverted. If we use the threshold, easy days can't work because easy days adjusts the card's interval within the fuzz range, which is very small.

user1823 (changes requested):

schedule/easy_days.py
```diff
@@ -22,27 +22,8 @@
...
-        if (
-            config.easy_days_review_ratio_list[
-                (sched_current_date() + timedelta(days=day_offset)).weekday()
-            ]
-            < 1
-            or (sched_current_date() + timedelta(days=day_offset)).strftime("%Y-%m-%d")
-            in config.easy_dates
-        ):
-            due_days.append(today + day_offset)
+    due_days = [today + day_offset for day_offset in range(35)]
```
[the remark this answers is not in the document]
user1823: Ok, that's perfectly reasonable.

schedule/reschedule.py
L-M-Sherlock: Done.

schedule/reschedule.py
L-M-Sherlock: Done.

schedule/easy_days.py
```diff
@@ -22,27 +22,8 @@
 def easy_days(did):
     config = Config()
     config.load()
-    if not config.load_balance:
-        tooltip("Please enable load balance first")
-        return
-    if (
-        all([r == 1 for r in config.easy_days_review_ratio_list])
```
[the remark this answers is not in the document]
user1823: Ok, makes sense

user1823 (approved):
Code-wise LGTM. Please ask someone else to verify if it works correctly for them. (I don't use Easy Days.)

Expertium:
I don't use Easy Days either

PR #503, opened 2024-11-30
Feat/Disperse siblings among cards rescheduled on another device

user1823:
Essentially, a revert of https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/434. We can now distinguish between "Set Due Date" and "Reschedule on change".

This is useful even if Desktop Anki < 24.11 because the other device may have Anki 24.11+.

### DIFF
--- a/sync_hook.py
+++ b/sync_hook.py
@@ -23,10 +23,10 @@ def review_cid_remote(local_rids: List[int]):
             f"""SELECT DISTINCT cid
             FROM revlog
             WHERE id NOT IN {local_rid_string}
-            AND ease > 0
+            {"AND type != 4" if config.auto_disperse_after_reschedule else "AND ease > 0"}
             AND (type < 3 OR factor != 0)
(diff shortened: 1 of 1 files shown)


L-M-Sherlock:
> This is useful even if Desktop Anki < 24.11 because the other device may have Anki 24.11+.

I don't want to maintain two versions.

L-M-Sherlock (approved):
LGTM

user1823:
> I don't want to maintain two versions.

Oh, yes. I just wanted to say that you can merge this PR before creating a separate branch for 23.10 to 24.06.3 so that both the versions would have this change. You won't need to do any further maintenance in the new branch.

L-M-Sherlock:
> This is useful even if Desktop Anki < 24.11 because the other device may have Anki 24.11+.

I don't recommend this use case, so I decide not to include this PR in old version.

PR #504, opened 2024-11-30
Feat/skip manually scheduled cards when rescheduling

user1823:
Because we can now distinguish between "Set Due Date" and "Reschedule on change", we don't need to add a separate configurable option.

This PR should be merged into main only after a separate branch is created for 23.10-24.06.3


L-M-Sherlock (approved):
LGTM

PR #505, opened 2024-12-02
Remove load from FSRS stats

user1823:
This PR removes the unused code that was previously used to calculate load. It should improve performance slightly because the add-on wouldn't have to calculate the load.


L-M-Sherlock (approved):
LGTM

PR #506, opened 2024-12-02
use `write_custom_data` to modify the value of `v`

ishiko732:
I need to store some data in `custom_data`, but `reschedule` overwrites my data. I want to clear only the value of the `v` field.


L-M-Sherlock (approved):
LGTM

PR #508, opened 2024-12-03
Feat/recommended steps

L-M-Sherlock:
Patch: [fsrs4anki-helper.ankiaddon.zip](https://github.com/user-attachments/files/17988513/fsrs4anki-helper.ankiaddon.zip)

Preview:

image

### DIFF
--- a/stats.py
+++ b/stats.py
@@ -187,7 +187,76 @@ def get_steps_stats(self: CollectionStats):
                 <td class="trr">{stats['retention']}</td>
                 <td class="trr">{format_time(results['stability'][rating])}</td>
                 <td class="trr">{stats['count']}</td>
-            </tr>"""
+            </tr>
+            """
+
+    html += f"""
+    <tr>
+        <td colspan="11" class="trl">
+            <strong>Desired retention:</strong>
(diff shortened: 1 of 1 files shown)


Expertium:
Suggestion: if recommended learning steps are <10s, just display "Not needed" or something like that

L-M-Sherlock:
> Also, why does it sometimes suggest one learning step, and sometimes two?

If the 2nd learning step is longer than the cutoff, it will be hidden.

user1823:
> Suggestion: if recommended learning steps are <10s, just display "Not needed" or something like that

Why?

user1823 (commented):
Functionally, LGTM.

Just being curious. Why is the code written in JS rather than Python?

Expertium:
> Why?

Because <10s is useless

user1823:
> Because <10s is useless

In that case, instead of suggesting the user to remove the steps completely, shouldn't we suggest them a slightly longer but more reasonable step (e.g., 10s or 15s)?

L-M-Sherlock:
> Just being curious. Why is the code written in JS rather than Python?

It's not written by me. I guess only JS could do such a reactive interaction in this page.

PR #510, opened 2024-12-04
Remove "Auto apply Easy Days"

user1823:
This feature is no longer required because Easy Days is now natively integrated in Anki.


L-M-Sherlock (approved):
LGTM

PR #514, opened 2024-12-05
Feat/Steps Stats for Good Then Again

L-M-Sherlock:
patch: [fsrs4anki-helper.ankiaddon.zip](https://github.com/user-attachments/files/18024545/fsrs4anki-helper.ankiaddon.zip)

preview:

image

Color suggestion wanted.

### DIFF
--- a/schedule/reschedule.py
+++ b/schedule/reschedule.py
@@ -66,9 +66,9 @@ class FSRS:
             preset_id = self.DM.config_dict_for_deck_id(did)["id"]
             self.due_cnt_per_day_per_preset[preset_id][due_date] += count
             self.did_to_preset_id[did] = preset_id
-            self.preset_id_to_easy_days_percentages[preset_id] = (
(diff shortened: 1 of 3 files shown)


aleksejrs:
"Good Then Again" and "Lapse" look distinct from the rest of the table's lines.

aleksejrs:
How about coloring the words in them like they are colored in their own rows? If needed, consider giving the "Then" regular weight.

aleksejrs:
But I haven't even thought about what it means. I guess "Good Then Again" is for the stability after "Again" that happened after "Good".

L-M-Sherlock:
> I guess "Good Then Again" is for the stability after "Again" that happened after "Good".

"Good Then Again" is for the stability where you firstly rate "Good" then rate "Again" in a new card.

L-M-Sherlock:
image

How about that?

aleksejrs:
Looks OK.

PR #515, opened 2024-12-06
add State column

L-M-Sherlock:
make it more clear

image


sommerluk:
Thanks, that helped me a lot to understand it!

PR #524, opened 2025-01-02
Fix/only auto dispersed rescheduled cards' siblings

### LINKED ISSUE #522: [BUG] After rescheduling, siblings of skipped cards shouldn't be dispersed.
If I use a non-zero value of `reschedule_threshold` and enable `auto_disperse_after_reschedule`, then I frequently get situations like "7 cards rescheduled. 600 cards in 300 notes dispersed."

The problem is that this makes `reschedule_threshold` useless because disperse reschedules the card anyway.

Solution:
If a card is skipped during reschedule, it should be excluded from auto-disperse.

To do this, instead of getting nid's here:
https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/8a1e7cc21b39ddc5006a368a095a5db07f934a52/schedule/reschedule.py#L306

we need to get the nid's of the rescheduled cards, somewhere after this:
https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/8a1e7cc21b39ddc5006a368a095a5db07f934a52/schedule/reschedule.py#L361

L-M-Sherlock:
fix #522 

patch: [fsrs4anki-helper.ankiaddon.zip](https://github.com/user-attachments/files/18289092/fsrs4anki-helper.ankiaddon.zip)

### DIFF
--- a/schedule/reschedule.py
+++ b/schedule/reschedule.py
@@ -348,8 +348,9 @@ def reschedule_background(
     cnt = 0
     cancelled = False
     rescheduled_cards = []
+    filtered_nids = set()
     undo_entry = mw.col.add_custom_undo_entry("Reschedule")
-    for cid, did, _, desired_retention, maximum_interval in cards:
+    for cid, did, nid, desired_retention, maximum_interval in cards:
         if cancelled:
             break
         fsrs.desired_retention = desired_retention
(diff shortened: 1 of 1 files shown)


user1823 (approved):
LGTM. Thanks!

PR #531, opened 2025-01-14
Fix/display not enough data instead of You don't need steps

L-M-Sherlock:
It solves the problem mention in https://www.reddit.com/r/Anki/comments/1i0glw6/when_fsrs_helper_says_that_you_dont_need_a/

Preview:

image

Patch:

[fsrs4anki-helper.ankiaddon.zip](https://github.com/user-attachments/files/18405096/fsrs4anki-helper.ankiaddon.zip)

### DIFF
--- a/stats.py
+++ b/stats.py
@@ -283,7 +283,7 @@ def get_steps_stats(self: CollectionStats):
                 learningStepRow.innerText = 'insufficient data to recommend the second step, please keep current setting';
             }} else {{
                 learningStepRow.innerText = (!learningStep1 && !learningStep2) 
-                    ? "You don't need learning steps" 
+                    ? "Keep the steps field blank." 
                     : `${{learningStep1}} ${{learningStep2}}`;
             }}
 
--- a/stats.py
+++ b/stats.py
@@ -257,12 +257,10 @@ def get_steps_stats(self: CollectionStats):
             return 1 / FACTOR * (Math.pow(dr, (1 / DECAY)) - 1);
         }}
 
-        function calculateStep(stability, factor, count) {{
+        function calculateStep(stability, factor) {{
             const step = stability * factor;
             if ((step >= cutoff || Number.isNaN(step))) {{
                 return '';
-            }} else if (count < 100) {{
-                return '(data is not enough, please keep current setting)';
             }}
             return formatTime(Math.max(step, 1));
         }};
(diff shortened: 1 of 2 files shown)


user1823:
This is what I got. I assume that the blank field is unexpected.

image

L-M-Sherlock:
Weird. I don't know how to reproduce it. Could you share the affected deck with me?

user1823:
[Default+.apkg.zip](https://github.com/user-attachments/files/18407898/Default%2B.apkg.zip) (rename to remove .zip)

L-M-Sherlock:
Fixed

patch: [fsrs4anki-helper.ankiaddon.zip](https://github.com/user-attachments/files/18408014/fsrs4anki-helper.ankiaddon.zip)

user1823:
This repetition also doesn't look good.
image

Also, in this case, the relearning steps field should also say "(not enough data)" because there are only 41 reviews.

L-M-Sherlock:
> This repetition also doesn't look good.

The problem is there are two learning steps. It's possible that the first one has enough data but the second one doesn't. Do you have any idea?

L-M-Sherlock:
image

[fsrs4anki-helper.ankiaddon.zip](https://github.com/user-attachments/files/18418186/fsrs4anki-helper.ankiaddon.zip)

cid: 1734102032614, 1734110579932

image

image

stats.py
```diff
@@ -272,13 +270,21 @@ def get_steps_stats(self: CollectionStats):
...
-
-            learningStepRow.innerText = (!learningStep1 && !learningStep2) 
-                ? "You don't need learning steps" 
-                : `${{learningStep1}} ${{learningStep2}}`;
+            const learningStep2 = calculateStep(Math.min(stability[2] * 2 - stability[1], stability[3], stability[4]), factor);
+
+            if (learningStep1Count < 100 && learningStep2Count < 100) {{
+                learningStepRow.innerText = '(data is not enough, please keep current setting)';
+            }} else if (learningStep1Count < 100) {{
+                learningStepRow.innerText = 'insufficient data to recommend the first step, please keep current setting';
```
user1823, line 281: If there is enough data to recommend one step but not both, this message appears and no step is recommended.

But IMO, we should recommend the step for which sufficient data is available.

stats.py
```diff
@@ -272,13 +270,21 @@ def get_steps_stats(self: CollectionStats):
...
+            }} else if (learningStep2Count < 100) {{
+                learningStepRow.innerText = 'insufficient data to recommend the second step, please keep current setting';
+            }} else {{
+                learningStepRow.innerText = (!learningStep1 && !learningStep2) 
+                    ? "You don't need learning steps"
```
user1823, line 286:
```suggestion
                    ? "Keep the steps field blank." 
```

stats.py
L-M-Sherlock: But how to display the message and the step simultaneously clearly? 同时

user1823: I was thinking something like `15m (insufficient data to recommend the second step, please keep current setting)`. But, it's probably unclear. Maybe @Expertium has a better idea?

Expertium: I think just "insufficient data" is fine, it gets the point across.
If there isn't enough data for any steps, display "insufficient data".
If there is sufficient data to recommend the first step but not the second step, then display only the first one, without "insufficient data".

PR #539, opened 2025-03-21
Expt/leech-detector

L-M-Sherlock:
modified from https://github.com/rbrownwsws/leechkit

patch: [fsrs4anki-helper.ankiaddon.zip](https://github.com/user-attachments/files/19385749/fsrs4anki-helper.ankiaddon.zip)

### DIFF
--- a/browser/detector.py
+++ b/browser/detector.py
@@ -208,7 +208,7 @@ def card_is_leech(
 
     if leech_threshold < 0 or leech_threshold > 1:
         raise Exception("leech_threshold must be between 0 and 1")
-    
+
     reviews.reverse()
     filtered_reviews = filter_out_reviews_unwanted_by_fsrs(reviews)
     grouped_reviews = group_card_reviews_by_day(
--- a/browser/detector.py
+++ b/browser/detector.py
@@ -204,10 +204,10 @@ def card_is_leech(
     next_day_starts_at_hour: int,
 ) -> (bool, Optional[float], Optional[float]):
     if skip_reviews < 1:
-        raise Exception("skip_reviews must be at least 1")
+        raise ValueError("skip_reviews must be at least 1")
 
     if leech_threshold < 0 or leech_threshold > 1:
-        raise Exception("leech_threshold must be between 0 and 1")
+        raise ValueError("leech_threshold must be between 0 and 1")
 
     reviews.reverse()
     filtered_reviews = filter_out_reviews_unwanted_by_fsrs(reviews)
--- a/browser/detector.py
+++ b/browser/detector.py
@@ -1,0 +1,243 @@
+import math
+from typing import Sequence, Final, Optional, Callable
+from dataclasses import dataclass
+
+from anki.cards import Card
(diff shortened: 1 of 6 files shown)


JSchoreels:
Hey @L-M-Sherlock, after using the leechkit, I opened a PR on its side to also limit only the N latest review-days, because for very long card, I don't think that every user would like to see if a card was a leech 50 reviews ago but is OK for the past few 6 months :)

The PR : https://github.com/rbrownwsws/leechkit/pull/7

L-M-Sherlock:
This branch is for experimental use. And most code is not wrote by me. Maybe it's better to develop a standalone add-on for leech detector.

JSchoreels:
I get your point but at the same time it's also quite related to how FSRS change certain older legacy concepts like a leech, so I don't think it would be that problematic to have it here... But I 100% agree on the fact it's quite experimental so it's probably a bit too early to really merge it and made it available to everyone.

Right now my biggest concern is the fact that "leech" look at the whole history and thus can mark a lot of things as leech when in fact they might not be anymore.

By extension, I wonder if a "leech" should not be simply redefined as a "highly improbably direct bad streak of Again". I'll post it in the discord too to see a bit opinions on it

browser/detector.py
```diff
@@ -0,0 +1,243 @@
...
+    Calculate the corrected threshold for determining if a card is a leech.
+
+    :param alpha: The base threshold probability.
+
+    :param n: The number of trials / valid reviews.
+
+    :return: The corrected threshold value based on the given parameters.
+    """
+
+    return alpha / (_THRESHOLD_FACTOR * (n**2))
```
Expertium, line 41: Remove this dynamic threshold, it sucks. Just take the threshold and divide it by 1.3, because that's what works well in practice. The number of cards tagged as leeches tends to be roughly 1.3-1.5 times greater than what you would theoretically expect.

Also, add a second threshold that must be higher than the first one. A card's p must go above the *second* threshold to become a leech. Use the same correction factor of 1.3-1.5.

For example, the first threshold is 3% and the second threshold is 33%. The bigger the difference, the fewer cards will be jumping around from "leech" to "not a leech".

Expertium, line 41: Remove this dynamic threshold, it sucks. Just take the threshold and divide it by 1.3, because that's what works well in practice. The number of cards tagged as leeches tends to be roughly 1.3-1.5 times greater than what you would theoretically expect.

Also, add a second threshold that must be higher than the first one. A card's p must go above the *second* threshold to become a leech. Use the same correction factor of 1.3-1.5.

For example, the first threshold is 3% and the second threshold is 33%. The bigger the difference, the fewer cards will be jumping around from "leech" to "not a leech".

Expertium (changes requested):

browser/detector.py
L-M-Sherlock: As I mentioned before, this branch is only used to verify the feasibility of implementing the leech-detector in add-on. I won't maintain it or add any new features. And most of codes are not developed by me, so I don't know how to modify it without breaking anything.

PR #540, opened 2025-03-22
Feat/i18n

Luc-Mcgrady:
Requires a git sub-module.
after checkout:
```
git submodule init 
git submodule update 
```
(Not sure if using a submodule/library is too heavy a solution)

Based on the language the user has Anki set to.
(Example)
image

I only have it set up for `__init__.py` right now but presume it could work for the rest.

I also changed the English a small amount https://github.com/open-spaced-repetition/fsrs4anki-helper/commit/ac477a897dc1f6b915f32d58ec1729acb49a9456.

Is this something you'd like?


L-M-Sherlock:
It's great!

Luc-Mcgrady:
I think this is done. You may want to wait for the actual translations before merging though.

L-M-Sherlock (approved):
LGTM

PR #557, opened 2025-04-30
Add German translation

sommerluk:


sommerluk:
> Sind nicht-wörtliche Übersetzungen möglich?
Die sind meiner Erfahrung nach die Besten!

> Wenn ja, würde es sich ggf. anbieten bspw. Abstände zw. Geschwisterkarten vergrößern zu verwenden.
Vielen Dank für den Tipp! Habe ich jetzt angewendet.

sommerluk:
From my side, this is now ready.

@GithubAnon0000, would you like to take a look at the statistics page? I could share a screenshot to provide some context—only if you’re interested, no pressure.

sommerluk:
No, nothing specific apart from yet discussed translation of "disperse siblings". Maybe check if it's generally understandable?

grafik

grafik

(Yes, I know some parts are still in English. Those are strings that are not available for translation.)

sommerluk:
Thanks a lot for the review! Your changes have been applied. (I had done quite a lot a spelling lapses.)

> **# 1** "FSRS-Kartenstatus" is confusing. I think you're refering to "memory state", which would be the "(Arbeits-)Speicherzustand". Though without context, I cannot really help finding a better string. Could you maybe show me where this is being used?

It's used in wildly different contexts: 
grafik
grafik
grafik
grafik
grafik

It describes the three parameters R, S and D that are used by FRSR and that are individual properties of each card.

I had a couple of ideas, but each of my ideas seemed odd in one or two of these contexts. Therefore, I would stick to the current translation. While it's not perfectly clear, at least it makes clear that it's a property of the card, and does not convey ambiguous meaning.

> **# 2** "Diese Kennzahl kann verwendet werden, um verschiedene Lernende zu vergleichen." This seems odd. Can you actually compare multiple people with this metric?
> 
> Not a fault with your translation though, as the english string is the same.

I would consider this out of scope for this PR.

Thanks a lot, @GithubAnon0000

sommerluk:
@L-M-Sherlock This PR is ready now.

Luc-Mcgrady (approved):

L-M-Sherlock (approved):
Thanks for you all!

PR #561, opened 2025-05-04
I18n/Hard coded values

Luc-Mcgrady:
ref: 
- https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/560#issue-3037318430

@soleuniverse101 Are there any more problems besides the that need to be fixed?

### DIFF
--- a/schedule/advance.py
+++ b/schedule/advance.py
@@ -8,8 +8,7 @@ def get_desired_advance_cnt_with_response(safe_cnt, did):
     inquire_text = t("advance-inquire-text") + "\n"
     notification_text = (
         t(
-            "advance-notification-text",
-            deck_collection=t("deck" if did else "collection"),
+            "advance-notification-text" if "deck" else "collection",
             count=safe_cnt,
         )
         + "\n"
--- a/i18n.py
+++ b/i18n.py
@@ -46,8 +46,6 @@ i18n.add_translation(
     "retention", tr.statistics_true_retention_retention(), locale=locale
 )
 i18n.add_translation("cards", tr.browsing_cards(), locale=locale)
-i18n.add_translation("deck", tr.decks_deck(), locale=locale)
-i18n.add_translation("collection", tr.browsing_whole_collection(), locale=locale)
 
 
 def t(*args, **kwargs):
--- a/locale/en_US.json
+++ b/locale/en_US.json
@@ -89,7 +89,8 @@
     "reviews-on-cards": "Reviews on cards",
 
     "advance-inquire-text": "Enter the number of cards to be advanced.\n",
-    "advance-notification-text": "For this %{deck_collection}, it is relatively safe to advance up to %{count} cards.\n",
+    "advance-notification-text-deck": "For this deck, it is relatively safe to advance up to %{count} cards.\n",
+    "advance-notification-text-collection": "For this collection, it is relatively safe to advance up to %{count} cards.\n",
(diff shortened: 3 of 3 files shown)


schedule/advance.py
```diff
@@ -8,8 +8,7 @@ def get_desired_advance_cnt_with_response(safe_cnt, did):
     inquire_text = t("advance-inquire-text") + "\n"
     notification_text = (
         t(
-            "advance-notification-text",
-            deck_collection=t("deck" if did else "collection"),
+            "advance-notification-text" if "deck" else "collection",
```
soleuniverse101, line 11: Shouldn't be the line be something like this ?
```py
"advance-notification-text-deck" if did else "advance-notification-text-collection"
```
I think that currently because strings evaluate to `true` this will always use "advance-notification-text" and will never depend on `did` like the precedent code

soleuniverse101:
> ref:
> 
> * [Add French translation #560 (comment)](https://github.com/open-spaced-repetition/fsrs4anki-helper/pull/560#issue-3037318430)
> 
> @soleuniverse101 Are there any more problems besides that which need to be fixed?

I think you covered it, thanks!

I didn't notice anything like this except for when the `%{deck_collection}` is used.

soleuniverse101:
Unfortunately the PR is blocked because the workflow finds the locale strings invalid after we remove `"advance-notification-text"`, and currently `"advance-notification-text-deck"` and `"advance-notification-text-collection"` are not implemented.

I can add it in French in my original PR and merge it before your PR but the following languages also need the change in German and Chinese and I'm not sure how to do it.

In Greek, the locale can be easily changed because `"advance-notification-text"` is not translated yet and there's just the English version : Greek. So we just need to replace it the same way you did in `en_US.json`

Lastly in British English, Japanese and Brazilian Portuguese, the key is not present so no change has to be made.

schedule/advance.py
Luc-Mcgrady: Oops :joy:

Luc-Mcgrady:
> Unfortunately the PR is blocked because the workflow finds the locale strings invalid after we remove "advance-notification-text", and currently "advance-notification-text-deck" and "advance-notification-text-collection" are not implemented.

> I can add it in French in my original PR and merge it before your PR but the following languages also need the change in German and Chinese and I'm not sure how to do it.

I was just making sure that there was no other values I need to change before merging this. The hard coded values are from Anki's translation so I can copy them in from there and paste them in which I assume won't be a problem since no one else has complained so far.

soleuniverse101:
> > Unfortunately the PR is blocked because the workflow finds the locale strings invalid after we remove "advance-notification-text", and currently "advance-notification-text-deck" and "advance-notification-text-collection" are not implemented.
> 
> > I can add it in French in my original PR and merge it before your PR but the following languages also need the change in German and Chinese and I'm not sure how to do it.
> 
> I was just making sure that there was no other values I need to change before merging this. The hard coded values are from Anki's translation so I can copy them in from there and paste them in which I assume won't be a problem since no one else has complained so far.

Then should I wait for your PR before adding the new locale strings in mine ?

Luc-Mcgrady:
> Then should I wait for your PR before adding the new locale strings in mine ?
you can add them and then the tests will work when the changes are merged in.

soleuniverse101:
> > Then should I wait for your PR before adding the new locale strings in mine ?
> 
> you can add them and then the tests will work when the changes are merged in.

done 👍

L-M-Sherlock (approved):
LGTM

PR #566, opened 2025-05-12
Feat/Dataset Export for Research

L-M-Sherlock:
It is helpful if we want to analyze certain user's collection. The exported `*.revlog` file could be converted to `parquet` files like this: https://github.com/open-spaced-repetition/anki-revlogs-dataset-builder/blob/main/build_parquet.py

## Changes
- Added a new menu item "Export Dataset for Research" under the helper menu
- Implemented `export_dataset()` function to export user's review logs
- Added corresponding translation string in English locale
- Created a dedicated `user_files` directory to store exported datasets

## Implementation Details
- The exported dataset is saved in the addon's `user_files` directory
- The filename is based on the user's profile name with `.revlog` extension
- The feature is accessible through the helper menu alongside other options

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -1,7 +1,8 @@
+from pathlib import Path
 from aqt.gui_hooks import deck_browser_will_show_options_menu, state_did_change
 from aqt import mw
 from aqt.qt import QAction
-from aqt.utils import tooltip, openLink, askUser
+from aqt.utils import tooltip, openLink, askUser, showInfo
 from typing import Callable
 
 from .dsr_state import init_dsr_status_hook
@@ -211,6 +212,17 @@ def sponsor(did=None):
 menu_sponsor = build_action(sponsor, t("sponsor-author"))
 
 
+def export_dataset(did=None):
+    addon = mw.addonManager.addonFromModule(__name__)
+    user_files = Path(mw.addonManager.addonsFolder(addon)) / "user_files"
+    user_files.mkdir(parents=True, exist_ok=True)
+    mw.col.export_dataset_for_research(f"{user_files}/{mw.pm.name}.revlog")
+    showInfo(f"Your dataset has been exported to {user_files}/{mw.pm.name}.revlog")
+
+
+menu_export_dataset = build_action(export_dataset, t("export-dataset"))
+
+
 def pass_fail(did=None):
     openLink("https://ankiweb.net/shared/info/876946123")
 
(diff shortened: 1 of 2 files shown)


__init__.py
```diff
@@ -1,7 +1,8 @@
+from pathlib import Path
 from aqt.gui_hooks import deck_browser_will_show_options_menu, state_did_change
 from aqt import mw
 from aqt.qt import QAction
```
Luc-Mcgrady, line 4:
```suggestion
from aqt.qt import QAction, QDesktopServices, QUrl
```

__init__.py
```diff
@@ -1,7 +1,8 @@
...
-from aqt.utils import tooltip, openLink, askUser
+from aqt.utils import tooltip, openLink, askUser, showInfo
```
Luc-Mcgrady, line 5:
```suggestion
from aqt.utils import openLink, askUser
```

__init__.py
```diff
@@ -211,6 +212,17 @@ def sponsor(did=None):
 menu_sponsor = build_action(sponsor, t("sponsor-author"))
 
 
+def export_dataset(did=None):
+    addon = mw.addonManager.addonFromModule(__name__)
+    user_files = Path(mw.addonManager.addonsFolder(addon)) / "user_files"
+    user_files.mkdir(parents=True, exist_ok=True)
+    mw.col.export_dataset_for_research(f"{user_files}/{mw.pm.name}.revlog")
+    showInfo(f"Your dataset has been exported to {user_files}/{mw.pm.name}.revlog")
```
Luc-Mcgrady, line 220:
```suggestion
    QDesktopServices.openUrl(QUrl.fromLocalFile(str(user_files.absolute())))
```

L-M-Sherlock:
> So, in the exported file, we won't be able to know when the user reset a card.

It's not necessary to know it via entry with ease = 0. The the state of first entry after resetting is learning.

This code removes the revlog before resetting:

https://github.com/open-spaced-repetition/anki-revlogs-dataset-builder/blob/d154016ec279b4c51b2680f643b31b852b92293f/build_parquet.py#L93-L104

user1823:
Oh yes, I overlooked that.

But, some users have abnormal revlog histories (e.g. they use Set Due Date after Reset, they used Reset while the card was still in Learning phase). So, if we keep the Reset entries in the dataset, analysis of the data to find problems would become easier.

L-M-Sherlock:
It requires some changes on Anki's codebase. And it will break the previous workflow of data processing.

user1823:
Yeah, but both codes will require only minor changes, IMO.

L-M-Sherlock:
OK. I will add a filter on the builder's workflow after you submit the PR.

PR #574, opened 2025-06-14
Ignore reschedule_threshold if called by auto_reschedule

user1823:
If the card is being rescheduled by auto_reschedule after sync, there is no reason to not reschedule it even if the interval didn't change much. The purpose of reschedule_threshold is to avoid backlogs. Backlog won't be created by rescheduling cards that were studied today on a different device.


L-M-Sherlock:
bugbot run

L-M-Sherlock (approved):
LGTM

PR #582, opened 2025-07-05
Fix/fill in the value of decay during reschedule if recompute=True

L-M-Sherlock:
Patch: [fsrs4anki-helper.ankiaddon.zip](https://github.com/user-attachments/files/21072413/fsrs4anki-helper.ankiaddon.zip)


user1823 (approved):
LGTM. Thanks.

PR #594, opened 2025-07-16
Update de_DE.json

sommerluk:


locale/de_DE.json
```diff
@@ -43,7 +43,7 @@
     "recommended-addons": "Empfohlene Erweiterungen",
 
     "step-stats": "Lernstufen",
-    "step-stats-disabled": "If you would like to see your (re)learning step statistics, you can enable them with <code>Tools > FSRS Helper > Show step stats.</code><br/><br/>This graph will slow down the speed at which this page loads if you have a lot of reviews, so enable it with caution.",
+    "step-stats-disabled": "Die Anzeige der Lernstufen-Statistik kann unter <code>Extras → FSRS-Helfer → Lernstufen-Statistik anzeigen</code> aktiviert werden. Bei umfangreichen Daten wird die Statistik-Seite langsamer laden. Aktivieren Sie diese Funktion nur bei Bedarf.",
```
L-M-Sherlock, line 46: The line break is missing. Consider to add one?

sommerluk: If there is no technical reason to add it (maybe overflow over the window), I would prefer not to. The way this German translation is done, I think it's easier understandable if kept together.

PR #603, opened 2025-08-26
Match Anki's load balancer and slightly improve performance

user1823:
Before the second commit:
```
1629 cards rescheduled, 896 cards in 448 notes dispersed in 0.87 seconds
```

After the second commit:
```
1629 cards rescheduled, 896 cards in 448 notes dispersed in 0.85 seconds
```


L-M-Sherlock (approved):
LGTM

PR #616, opened 2025-09-18
Feat/Option to allow rescheduling of manually scheduled cards

Luc-Mcgrady:
partially reverts #504

ref:
- https://www.reddit.com/r/Anki/comments/1njwhdp/fsrs_ignoring_manually_rescheduled_cards_how_can/

c.c. @user1823


user1823:
What if we call this "Include manually scheduled cards when rescheduling" and default to `False` to emphasize that excluding them is the more reasonable choice?

Also, what if we don't add a menu option and make it configurable only via the add-on config so that people don't enable it without giving it a proper thought?
And if we are adding a menu option, what about adding a warning that says something like "Enabling this will cause reschedule to undo manually set due dates."?

__init__.py
```diff
@@ -113,6 +113,16 @@ def set_auto_disperse_after_reschedule(checked, _):
 )
 
 
+def set_skip_manual_resched_cards(checked, _):
```
user1823, line 116: This function should also be renamed.

L-M-Sherlock (approved):
LGTM. Thanks!

PR #621, opened 2025-09-27
Fix/Ensure fuzz doesn't go backward during rescheduling or dispersing siblings

### LINKED ISSUE #620: Use last interval instead of elapsed days for lower limit of fuzz
To calculate the lower limit of fuzz, the addon uses the elapsed_days. For example,

https://github.com/open-spaced-repetition/fsrs4anki-helper/blob/a295c3ddca43c48e63d5b4d352089578ead07dc9/schedule/disperse_siblings.py#L107-L110

However, elapsed days was just a compromise that we made and I later realized that we can get the actual value of the previous interval from the revlog.

I tried the following. But, it gave an error. I believe that I am making a trivial mistake here. Can you help me fix this?

```py
    last_interval = revlogs[0].lastIvl if len(revlogs) >= 1 else 0
    min_ivl, max_ivl = get_fuzz_range(new_ivl, last_interval, maximum_interval)
```

In the above code, I want to get the value of `lastIvl` column from the latest revlog entry.

L-M-Sherlock:
close #620

### DIFF
--- a/dsr_state.py
+++ b/dsr_state.py
@@ -2,7 +2,12 @@ from anki import hooks
 from anki.template import TemplateRenderContext, TemplateRenderOutput
 
 from .configuration import Config
-from .utils import get_decay, power_forgetting_curve, get_last_review_date, mw
+from .utils import (
+    get_decay,
+    power_forgetting_curve,
+    get_last_review_date_and_interval,
+    mw,
+)
 from .i18n import t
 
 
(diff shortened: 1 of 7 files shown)


user1823:
This causes an error:
```py
AttributeError: 'Card' object has no attribute 'lastIvl'
```
which makes sense because lastIvl is a property of the revlog, not the card.

L-M-Sherlock:
It should work with https://github.com/ankitects/anki/pull/4365

user1823:
This will be a regression for disperse_siblings for all the current users because the current version uses last_elapsed_days, while this PR will use 0 (because no release of Anki exports last_interval yet).

To fix this, if the `getattr` fails, we can use
```py
    last_interval = (
        int((revlogs[0].time - revlogs[1].time) / 86400) if len(revlogs) >= 2 else 0
    )
```

Also, we should rename `elapsed_days` in `get_fuzz_range` to `last_interval`.

PR #624, opened 2025-10-14
Respect Anki's Load Balancer Configuration 

JSchoreels:
When the `mw.col._set_load_balancer_enabled(False)` is set, now the Addon will skip the Fuzz part.

I kept some of the logic unchanged though, please feel free to say if you think the property should also alter some of those features

For example this part that I don't really understand what the point is

```
        if fsrs.reschedule_threshold > 0 and not (
            fsrs.apply_easy_days or auto_reschedule
        ):
            dr = fsrs.desired_retention
            odds = dr / (1 - dr)

            odds_lower = (1 - fsrs.reschedule_threshold) * odds
            dr_lower = odds_lower / (odds_lower + 1)
            adjusted_ivl_upper = next_interval(s, dr_lower, -decay)

            odds_upper = (1 + fsrs.reschedule_threshold) * odds
            dr_upper = odds_upper / (odds_upper + 1)
            adjusted_ivl_lower = next_interval(s, dr_upper, -decay)

            if card.ivl >= adjusted_ivl_lower and card.ivl <= adjusted_ivl_upper:
                if recompute:
                    return card, False
                else:
                    return None, False
```

CleanShot 2025-10-14 at 15 18 21@2x

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 __pycache__
 *.ipynb
 meta.json
+*.ankiaddon


L-M-Sherlock (approved):
LGTM, thanks!

PR #626, opened 2025-10-17
Use Anki's built-in fuzz if load balancer disabled

user1823:


L-M-Sherlock:
After thinking about it, I find that the current PR is consistent with Anki's behavior. I will merge it soon.

@JSchoreels, if you want to disable fuzz completely, please consider to submit a new PR.

JSchoreels:
> Anki's fuzz is deterministic. So, if you reschedule multiple times, the outcome won't change.

In the doc Fuzz is said to be randomly generated : https://docs.ankiweb.net/studying.html#fuzz-factor

I can't find the exact line where the randomness is inserted in the Anki project, so I'm a bit confused between the doc and your comment/the code.

schedule/reschedule.py
```diff
@@ -201,6 +201,10 @@ def load_balance(
     def apply_fuzz(self, ivl):
         if ivl < 2.5:
             return ivl
+
+        if not self.load_balancer_enabled:
+            return ivl + mw.col.fuzz_delta(self.card.id, ivl)
```
L-M-Sherlock: It doesn't respect maximum_interval.

L-M-Sherlock: Oh, wait. It respects. Nvm.

user1823: Are you sure?

https://github.com/ankitects/anki/blob/321e23acb2b02d657ca08727546c016d7ef27381/rslib/src/scheduler/states/fuzz.rs#L47-L63

Edit: I posted this before seeing your comment that it respects.

PR #631, opened 2025-10-23
Perf: Avoid unnecessary load balancer initialization

user1823:
We don't need to initialize the LB if it is disabled in Anki and specific due date isn't set


L-M-Sherlock (approved):
LGTM

PR #652, opened 2026-05-03
Feat: Push rescheduled cards after auto-reschedule/disperse

user1823:
Automatically push the auto-rescheduled and auto-dispersed cards to decrease the chances of sync conflicts when the same cards are reviewed on another device.

Fixes #648
Supersedes and closes #649

FYI @chrislongros

PR #654, opened 2026-06-08
Fix: skip filtered-deck cards in post-sync autoreschedule/disperse

chrislongros:
Follow-up to #652.

If a card is in a filtered deck (`odid != 0`), the post-sync reschedule only updates its `odue` and the card stays in the filtered deck — but it still updates `mod`/`usn`. The end result is that reviews on other devices get overwritten.

Fix: skip cards that are in a filtered deck. The card keeps the schedule from its review and is rescheduled normally once it's back in its home deck.


L-M-Sherlock (approved):
LGTM

